PR #2, opened 2020-06-18
python print to javascript console log using generic AST

jackdavidweber:
for code snippett use


jackdavidweber:
See genericTemplate.json which will serve as a continually updated marker of our progress and will hopefully ensure that the generics work across languages

PR #4, opened 2020-06-18
Arguments

jackdavidweber:
Built out some python ast to generic ast functionality so that it can handle arguments!

Would really appreciate feedback and your thoughts

### DIFF
--- a/experimental/genericTemplate.json
+++ b/experimental/genericTemplate.json
@@ -7,7 +7,7 @@
             "body": [
                 {
                     "type": "logStatement",
-                    "attributes": []
+                    "args": []
                 }
             ]   
         }
@@ -20,9 +20,19 @@
             "body": [
                 {
                     "type": "logStatement",
-                    "attributes": ["hello world"]
+                    "args": ["hello world"]
                 }
             ]   
+        }
+    },
+    "customStatementNoArgs": {
+        "py": "custom_function_name()",
+        "js": "customFunctionName()",
+        "gast": {
+            "type": "root",
+            "body": [
+                {"type": "customStatement"}
+            ]
         }
     }
 }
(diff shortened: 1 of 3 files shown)


experimental/genericTemplate.json
```diff
@@ -20,9 +20,19 @@
...
                 {
                     "type": "logStatement",
-                    "attributes": ["hello world"]
+                    "args": ["hello world"]
                 }
             ]   
         }
+    },
+    "customStatementNoArgs": {
+        "py": "custom_function_name()",
```
Danish21, line 29: The way I was imagining it was that the generic tree has no knowlede about python or js, but that each language converts to that format and prints from that format. 

In this case you would store the statement name in some chosen casing maybe camel case and then python would convert to that and so would js. Then when going from gast -> python you  would again convert to the format of that language.

experimental/jackPython/astMapping/main.py
```diff
@@ -1,20 +1,69 @@
...
+    elif type(op) == ast.Div:
+        return "/"
+    elif type(op) == ast.Sub:
+        return "-"
+
+"""
+converts a python ast BinOp and converts it to a readable string recursively 
+"""
+def binOp_to_str(bop):
+    if type(bop.left) == ast.BinOp:
```
g-harel, line 22: Is it not possible for `bop.right` to also be a `BinOp`?

jackdavidweber: Not that I have seen, but I haven't done extensive testing yet.

experimental/genericTemplate.json
jackdavidweber: yeah- agreed! we were honestly just using the generic template as a way to coordinate efforts across python and javascript. That way, @swalsh15 would know how to format conversion from jsAst to gast in the same way that I do it from pyAst. The language specific stuff was mostly to make it more clear for us in case of ambiguity in the name of "customStatementNoArgs"

PR #8, opened 2020-06-19
Flatten if statements

jackdavidweber:
Re-wrote program so that it works by recursively calling itself and helper functions. No longer has very much nesting at all since it keeps on just calling itself.

Would appreciate feedback on whether this improves readability. I think it definitely makes it easier to scale since new features can be added without having to write much code.


experimental/jackPython/astMapping/main.py
```diff
@@ -116,11 +130,11 @@ def translator(input_filename, ds, inputLanguage, outputLanguage):
...
-        print(astToJsMap[node["type"]][outputLanguage])
+        print(ds[node["type"]][outputLanguage])
 
 
 #TODO: replace the mapping for each one with a function that returns the correct output string WITH arguments
-astToJsMap = {
+ds = {
     "logStatement": {
         "py": "print()",
         "js": "console.log()"
```
Danish21, line 140: IMO the python converter one should not have logic anything related to JS nor should the gast.

experimental/jackPython/astMapping/main.py
```diff
@@ -26,79 +26,93 @@ def binOp_to_str(bop):
...
-    
-    return gast
+def name(node):
+    if node.id == "print":
+        return "logStatement"
+    else:
+        return "customStatement"
+                
+
+def node_to_gast(node):
```
Danish21, line 86: Is this name accurate? It seems like this is printing the string representation and not creating a gast?

Danish21, line 86: In general I like this approach of classifying the top level and recursing. I believe this should work. Not sure about edge cases though. 

Soon we will want to start creating a test suite that we can start testing changes against.

jackdavidweber: I think that this may be a design flaw in the way i wrote the function. Basically, the function will return a gast if a node that contains sub-nodes is passed to it. If the function is passed a base case, then it will simply return the string representation of the node.

For example:

- `node_to_gast(ast_generated_from_print("hello")) `will return the gast associated with `print("hello")`
- `node_to_gast(ast.String(s="hello"))` will return the string "hello"

This works since python is so flexible, but could definitely be a flaw. Does this make sense? May be easier to explain over video.
@Danish21

experimental/jackPython/astMapping/main.py
```diff
@@ -26,79 +26,93 @@ def binOp_to_str(bop):
...
-        return arg.s
-    elif type(arg) == ast.Num:
-        return arg.n
-    elif type(arg) == ast.BinOp:
-        return binOp_to_str(arg)
-    else:
-        return ""
-
-
+def module(node):
```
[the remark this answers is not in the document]
jackdavidweber: done!

experimental/jackPython/astMapping/main.py
```diff
@@ -26,79 +26,93 @@ def binOp_to_str(bop):
...
+takes a node that represents a list of nodes.
+returns a list of gast
 """
-def pyargs_to_strlist(args):
-    out = []
-    for arg in args:
-        out.append(pyarg_to_str(arg))
-            
-    return out
+def node_list(node):
```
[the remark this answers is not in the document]
jackdavidweber: got it!

experimental/jackPython/astMapping/main.py
jackdavidweber: I just separated py_to_gast into its own separate file in order to solidify this distinction

experimental/jackPython/astMapping/main.py
jackdavidweber: I think that's my next step!

PR #12, opened 2020-06-22
Move python to translate

jackdavidweber:
We wanted to move all of the python code into our translate folder which would essentially serve as the library for all of the ast translation. We structured the folders in a way that would be pretty easy for us to divide the work up between all contributors without causing merge issues.

The next step was to simply break up the python to gast files we had been working on into the folder structure we created. This ended up being much more difficult to do because of import issues. Evidently, python does not make it easy to import across different files. Even once we figured out a workaround using `sys.path.insert` we still had errors. For some reason `import filename` was working but `from filename import *` was not working. 

This PR contains a pretty hacky solution that took way too long to get to but seems to work. Would love thoughts on more elegant ways of dealing with this although both Cory and I spent some time looking for easier solutions and couldn't find anything.


jackdavidweber:
@swalsh15 definitely make sure to check out how we did this when you start with javascript stuff

translate/routers/py_router.py
```diff
@@ -1,3 +1,14 @@
+import sys
+import ast
+sys.path.insert(1,'helpers')
+import jack_py_helpers as h
```
Danish21, line 4: Use more descriptive names than h and e for the imports to enhance readability at the call sites.

translate/assign/py_assign.py
```diff
@@ -0,0 +1,20 @@
...
+
+sys.path.insert(1,'routers')
+import py_router as pr
+
+"""
+takes python ast assigns and converts them to generic ast format
+note, that this assumes only a single assignment (i.e. x = 4)
+for now, it does not work for things link x,y = 4,5
+example:
+    exampleIn Assign(targets=[Name(id='x')], value=Num(n=5))
```
Danish21, line 11: Thanks for adding detailed function comments!

translate/helpers/jack_py_helpers.py
```diff
@@ -0,0 +1,66 @@
+import sys
```
Danish21, line 1: Are these helpers still experimental? If not, IMO this file should be called py_helpers.py to ensure that everyone reuses each other's code and don't reimplement things.

jackdavidweber:
@swalsh15 do you want us to merge this branch?

PR #14, opened 2020-06-23
Move java script to translate

swalsh15:
Moved code to create JS AST with variable declarations to translate folder.
Tried to standardize imports by finding path using os library.

Import errors were from pylint, we're planning to complete the backend of project on PyCharm. PyCharm linter underlines the import error when loaded. In order to make stop this underlining right-click on each folder and click Mark Directory As -> Source directory.


translate/var_declare/js_var_declare.py
```diff
@@ -0,0 +1,19 @@
+import sys
+import os
+
+# get path automatically and handle imports
+path = os.path.dirname(os.getcwd()) + '/helpers'
+sys.path.insert(1, path)
+import steph_helpers
```
jackdavidweber, line 7: probably should just make steph_helpers into js_helpers and then just have a separate py_helpers folder

PR #18, opened 2020-06-24
Add expressions to JS AST. Fixed imports in all files by adding paths…

swalsh15:
Added JS Expression function and standardized imports


swalsh15:
Added all paths for imports in js_main so imports work as expected in files in nested files. Added rest of JS AST functions to translate folder

PR #21, opened 2020-06-24
working gast to js for var assign and logstatements

jackdavidweber:
provided example that runs well at the bottom

### DIFF
--- a/translate/gast_to_js.py
+++ b/translate/gast_to_js.py
@@ -1,0 +1,65 @@
+
+
+def list_to_csv_str(l):
+    s = ""
+    for i in l:
+        s += str(i) + ", "
+    
+    return s[:-2] # remove last comma and space
+
+
+def logStatement(gast):
+    arg_string = list_to_csv_str(gast["args"])
+    return "console.log(" + arg_string + ")"
+
+
+def value_to_str(val):
+    if type(val) == str:
+        return '"' + val + '"'
+    else:
+        return str(val)
+
+
+def varAssign(gast):
+    value = value_to_str(gast["varValue"])
+    return "const " + gast["varId"] + " = " + value
+
+
+def body(gast_list):
+    out = ""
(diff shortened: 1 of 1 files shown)


translate/gast_to_js.py
```diff
@@ -0,0 +1,65 @@
+
+
+def list_to_csv_str(l):
+    s = ""
+    for i in l:
+        s += str(i) + ", "
+    
+    return s[:-2] # remove last comma and space
```
g-harel, line 8: Might be easier to use [join](https://www.tutorialspoint.com/python/string_join.htm) to avoid  this issue entirely.

translate/gast_to_js.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def logStatement(gast):
+    arg_string = list_to_csv_str(gast["args"])
+    return "console.log(" + arg_string + ")"
+
+
+def value_to_str(val):
+    if type(val) == str:
```
g-harel, line 17: Where is `str` coming from?

translate/gast_to_js.py
```diff
@@ -0,0 +1,65 @@
...
+        return '"' + val + '"'
+    else:
+        return str(val)
+
+
+def varAssign(gast):
+    value = value_to_str(gast["varValue"])
+    return "const " + gast["varId"] + " = " + value
```
g-harel, line 25: Not sure const is the right one here, variables may be reassigned lower down in the code.

translate/gast_to_js.py
jackdavidweber: stored in gast. For example x= "hello" would store "hello" in gast which would then have to be converted out.

translate/gast_to_js.py
jackdavidweber: This is a place-holder for now- i will add a TODO

PR #23, opened 2020-06-24
Create python AST from string instead of file

swalsh15:
Use ast library instead of astor library to create python AST so we can use a string instead of file to build. Still works as expected and can print a "pretty" version of AST using astor.dump_tree().

### DIFF
--- a/translate/py_main.py
+++ b/translate/py_main.py
@@ -10,7 +10,7 @@ sys.path.append('routers')
 import py_router
 
 """
-takes pyton file and converts it to a node
+takes pyton string and converts it to a node
 node is then dealt with by node_to_gast 
 """
 def py_to_gast(python_input):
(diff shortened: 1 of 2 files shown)


translate/py_main.py
```diff
@@ -13,11 +13,11 @@
...
-def py_to_gast(python_input_filename):
-    input_ast =  astor.code_to_ast.parse_file(python_input_filename)
+def py_to_gast(python_input):
+    input_ast = ast.parse(python_input)
     print(astor.dump_tree(input_ast))
     return py_router.node_to_gast(input_ast)
 
-fileName = "py_sample_code.py"
-gast = py_to_gast(fileName)
+file = "print('hello world')"
```
augerm, line 21: Can we rename this to `src_code_input` (or something similar) now that this isn't a file anymore?

augerm: `python_input` like in `py_to_gast` works

translate/py_main.py
```diff
@@ -13,11 +13,11 @@
 takes pyton file and converts it to a node
```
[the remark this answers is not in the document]
jackdavidweber: got it

translate/py_main.py
jackdavidweber: got it

PR #27, opened 2020-06-25
Add language choice functionality to POST request

jackdavidweber:
Before, POST request was only able to handle translations from js to py. By adding in_lang and out_lang in the POST request, it is able to translate btwn js and py as well as btwn py and js. This was only possible because of [work done](https://github.com/jackdavidweber/cjs_capstone/tree/stringToAST) by @swalsh15 to allow python to gast to use a string as an input

Here is an example of how the post request looks in action.
image

Note: I added in some logic so that the current frontend request to the backend should still work. I will deploy so that we can edit the request whenever.

### DIFF
--- a/app.py
+++ b/app.py
@@ -18,16 +18,29 @@ api = Api(app)
 CORS(app)
 
 parser = reqparse.RequestParser()
-parser.add_argument('input')
+parser.add_argument('in_code')
+parser.add_argument('in_lang')
+parser.add_argument('out_lang')
 
 class Translate(Resource):
     def get(self):
         return {'work': 'ing'}
 
     def post(self):
+        # bring in post arguments
         args = parser.parse_args()
-        input_text = args['input']
-        gast = js_main.js_to_gast(input_text)
+        input_code = args['in_code']
+        input_lang = args['in_lang']
+        output_lang = args['out_lang']
+
+        print(input_code, input_lang, output_lang)
+        if input_lang == "js":
+            gast = js_main.js_to_gast(input_code)
+        elif input_lang == "py":
+            pass # TODO: fill this in
+        else:
+            return {"Error": "must specify valid input language"}
+        
         output_code = gast_to_py.gast_to_py(gast)
         return {'response': output_code}
 
--- a/app.py
+++ b/app.py
@@ -10,6 +10,7 @@ sys.path.append('translate/helpers')
 sys.path.append('translate/routers')
 
 import js_main as js_main
+import py_main as py_main
 import gast_to_py as gast_to_py
 import gast_to_code as gtc
 
@@ -38,7 +39,7 @@ class Translate(Resource):
         if input_lang == "js":
             gast = js_main.js_to_gast(input_code)
         elif input_lang == "py":
-            pass # TODO: fill this in
+            gast = py_main.py_to_gast(input_code)
         else:
             return {"Error": "must specify valid input language"}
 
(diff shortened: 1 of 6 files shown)


jackdavidweber:
Just deployed- everything seems to work fine!

app.py
```diff
@@ -10,7 +10,9 @@
 sys.path.append('translate/routers')
 
 import js_main as js_main
```
Danish21, line 12: The alias for imports with the same name are redundant. Delete the alias on those cases.

app.py
```diff
@@ -19,16 +21,34 @@
...
+            input_lang = "js"
+            output_lang = "py"
+
+        if input_lang == "js":
+            gast = js_main.js_to_gast(input_code)
+        elif input_lang == "py":
+            gast = py_main.py_to_gast(input_code)
+        else:
+            return {"Error": "must specify valid input language"}
+
```
Danish21, line 49: Also do some error checking on the output lang like you do for the input.

app.py
```diff
@@ -19,16 +21,34 @@
...
+        if input_lang == None and output_lang == None:
+            input_lang = "js"
+            output_lang = "py"
+
+        if input_lang == "js":
+            gast = js_main.js_to_gast(input_code)
+        elif input_lang == "py":
+            gast = py_main.py_to_gast(input_code)
+        else:
+            return {"Error": "must specify valid input language"}
```
augerm, line 48: Might be helpful to specify in the response the valid input types.

app.py
```diff
@@ -19,16 +21,34 @@
...
+        # bring in post arguments
         args = parser.parse_args()
-        input_text = args['input']
-        gast = js_main.js_to_gast(input_text)
-        output_code = gast_to_py.gast_to_py(gast)
+        input_code = args['input']
+        input_lang = args['in_lang']
+        output_lang = args['out_lang']
+
+        # TODO: remove this once frontend is updated
```
g-harel, line 38: Even once the frontend is updated, requests shouldn't be assumed to be correct. This seems like a perfect situation to return an error code in the 400 range.

translate/expression/py_expression.py
```diff
@@ -26,10 +26,10 @@ def call(node):
...
+FIXME: this should prob be in helpers since it is also used by assign
 """
 def name(node):
     if node.id == "print":
         return "logStatement"
     else:
-        return "customStatement"
-      
\ No newline at end of file
+        return node.id
```
augerm, line 35: Shouldn't this still be a type?

If we want to return the ID it should probably be in an additional field that exists on `customStatement` objects.

g-harel, line 35: Out of curiosity, what's the fix here?

translate/py_main.py
```diff
@@ -10,14 +9,9 @@
 import py_router
 
 """
-takes pyton file and converts it to a node
+takes pyton string and converts it to a node
```
g-harel, line 12: Nit: `python code` might make it clearer it's not a `""` string.

app.py
jackdavidweber: got it

translate/expression/py_expression.py
jackdavidweber: I was storing customStatement instead of the actual name. It was a remnant of how we were planning on dealing with custom functions that I forgot to remove. This is another reason why we are going to need types in our gast

jackdavidweber: added a FIXME- i think this plays into the broader types issue that we are going to face

PR #30, opened 2020-06-25
Boolean and arrays support

cjoshea9:


cjoshea9:
Does not work perfectly for boolean and arrays going back and forth

cjoshea9:
We are going to hold off on this branch until we create a better contract for different types.

PR #32, opened 2020-06-26
Type restructure js to gast

swalsh15:
Literal values are now stored with primitive type in generic ast (num, bool, string). Added support for array declarations and binary and boolean operations. General AST now takes a nested format. Removed old iterative helper functions for arrays.

### DIFF
--- a/translate/helpers/js_helpers.py
+++ b/translate/helpers/js_helpers.py
@@ -17,6 +17,12 @@ def jsarg_to_str(arg):
   else:
     return ""
 
+def js_array_expression(node):
+    gast = {"type" : "arr"}
+    gast["elts"] = []
+    for elm in node.elements:
+        gast["elts"].append(js_router.node_to_gast(elm))
+    return gast
 """
 takes ast node of type program and returns
 a generic ast for that node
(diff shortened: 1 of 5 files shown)


translate/helpers/js_helpers.py
```diff
@@ -30,15 +36,39 @@ def program_to_gast(node):
...
+  return gast
+
+def boolOp(node):
+    gast = {"type": "boolOp"}
+    if node.left.type != "LogicalExpression" and node.right.type != "LogicalExpression":
+        gast["left"] = js_router.node_to_gast(node.left)
+        gast["op"] = node.operator
+        gast["right"] = js_router.node_to_gast(node.right)
+    elif node.left.type == "LogicalExpression":
+        gast["left"] = boolOp(node.left)
```
g-harel, line 64: Could the `boolOp` here be replaced with `node_to_gast` too? Then there would be no need to check for `"LogicalExpression"` in here too and it would get a lot simpler. Same thing above for `binOp`.

jackdavidweber: Need to work on merging so I am adding a todo linking to this comment in the code.

g-harel: Just as another data point, this is why I love having high test coverage so much. You can try  stuff out that you think might work and the tests will make sure all the things you care about haven't changed.

PR #35, opened 2020-06-26
Type restructure merge

jackdavidweber:
Final merge and work out of bugs on our big type_restructure undertaking. This PR is just the simple merging of three large PR's responsible for adding types to the gast as the interns decided upon:
- [type_restructure_py_to_gast](https://github.com/jackdavidweber/cjs_capstone/pull/33)
- [type_restructure_js_to_gast](https://github.com/jackdavidweber/cjs_capstone/pull/32)
- [type_restructure_gast_to_code](https://github.com/jackdavidweber/cjs_capstone/pull/34)

### DIFF
--- a/gast_to_code.py
+++ b/gast_to_code.py
@@ -1,45 +1,51 @@
-
-# general_helpers
-def value_to_str(val):
-    if type(val) == str:
-        return '"' + val + '"'
-    else:
-        return str(val)
-
-def body(gast_list, out_lang):
+"""
+Helper for lists of gast
+Default is to put comma and space btwn each stringified gast
+    i.e. list_helper({str_gast}, {str_gast}, out_lang) --> str, str
+Can specify different btwn string with third parameter
+    i.e. list_helper({str_gast}, {str_gast}, out_lang, "**") --> str**str
+"""
+def list_helper(gast_list, out_lang, btwn_str = ", "):
     out = ""
     for gast in gast_list:
         out += gast_router(gast, out_lang)
-        out += "\n"
-
-    return out[:-1] # remove \nS
-
-def list_to_csv_str(l):
-    s = ""
-    for i in l:
-        if type(i) == str:
-            i = '"' + i + '"'
-        s += str(i) + ", "
+        out += btwn_str
     
-    return s[:-2] # remove last comma and space
-
+    num_btwn_chars = len(btwn_str)
+    return out[:-num_btwn_chars] # remove \nS
+        
 # py_specific_helpers
 def py_logStatement(gast):
-    arg_string = list_to_csv_str(gast["args"])
+    arg_string = gast_router(gast["args"],"py")
     return "print(" + arg_string + ")"
 
 def py_varAssign(gast):
-    value = value_to_str(gast["varValue"])
+    value = gast_router(gast["varValue"], "py")
     return gast["varId"] + " = " + value
 
 # js_specific_helpers
 def js_logStatement(gast):
-    arg_string = list_to_csv_str(gast["args"])
+    arg_string = gast_router(gast["args"],"py")
     return "console.log(" + arg_string + ")"
 
 def js_varAssign(gast):
-    value = value_to_str(gast["varValue"])
-    return "const " + gast["varId"] + " = " + value
+    kind = gast["kind"]
+    varId = gast["varId"]
+    varValue = gast_router(gast["varValue"], "js")
+    
+    return kind + " " + varId + " = " + varValue
+
+def py_bool(gast):
+    if gast["value"] == 1:
+        return "True"
+    else:
+        return "False"
+
+def js_bool(gast):
+    if gast["value"] == 1:
+        return "true"
+    else:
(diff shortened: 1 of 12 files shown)


gast_to_code.py
```diff
@@ -1,45 +1,51 @@
...
+    return kind + " " + varId + " = " + varValue
+
+def py_bool(gast):
+    if gast["value"] == 1:
+        return "True"
+    else:
+        return "False"
+
+def js_bool(gast):
+    if gast["value"] == 1:
```
g-harel, line 45: Any reason we're using numbers to represent booleans in the gAST? JSON supports booleans natively.

gast_to_code.py
```diff
@@ -61,11 +71,27 @@ def js_varAssign(gast):
...
-        return body(gast["body"], out_lang)
+    if type(gast) == list:
+        return list_helper(gast, out_lang)
+
+    # Primitives
+    elif gast["type"] == "num":
+        return str(gast["value"])
+    elif gast["type"] == "arr":
+        return "[" + gast_router(gast["elts"], out_lang) + "]"
+    elif gast["type"] == "str":
```
g-harel, line 82: For another time, but it may cause issues if multiline strings are not differentiated.

translate/assign/py_assign.py
```diff
@@ -11,6 +11,7 @@
 def assign(node):
     gast = {}
     gast["type"] = "varAssign"
+    gast["kind"] = "let"
```
g-harel, line 14: Is this the way non-constant variables are being annotated in the gAST? Seems pretty js-specific.

gast_to_code.py
```diff
@@ -1,45 +1,51 @@
...
-
-def body(gast_list, out_lang):
+"""
+Helper for lists of gast
+Default is to put comma and space btwn each stringified gast
+    i.e. list_helper({str_gast}, {str_gast}, out_lang) --> str, str
+Can specify different btwn string with third parameter
+    i.e. list_helper({str_gast}, {str_gast}, out_lang, "**") --> str**str
+"""
+def list_helper(gast_list, out_lang, btwn_str = ", "):
```
augerm, line 8: `btwn_str` -->  `csv_delimiter`

gast_to_code.py
```diff
@@ -1,45 +1,51 @@
...
-    for i in l:
-        if type(i) == str:
-            i = '"' + i + '"'
-        s += str(i) + ", "
+        out += btwn_str
     
-    return s[:-2] # remove last comma and space
-
+    num_btwn_chars = len(btwn_str)
+    return out[:-num_btwn_chars] # remove \nS
```
augerm, line 15: Comment should be `#remove last comma and space`

gast_to_code.py
```diff
@@ -1,45 +1,51 @@
...
-    s = ""
-    for i in l:
-        if type(i) == str:
-            i = '"' + i + '"'
-        s += str(i) + ", "
+        out += btwn_str
     
-    return s[:-2] # remove last comma and space
-
+    num_btwn_chars = len(btwn_str)
```
augerm, line 14: `csv_delimiter_len = ...`

augerm: Better would be to just `break` from the loop early instead of removing the string after the fact. You can grab the index in a for in loop using `enumerate()` and check if it's the last element. This logic is a little awkward as is.

https://stackoverflow.com/a/522578

gast_to_code.py
jackdavidweber: Not really- we just made that decision together since we kept on getting confused with capitalization. Also when converting back to code, 0 often is false and 1 is true which we liked. Do we need to change?

gast_to_code.py
jackdavidweber: @g-harel can you elaborate? Not quite sure what you mean

translate/assign/py_assign.py
jackdavidweber: That's the way we decided to do it across all languages to make things easier but agreed it does seem a bit "hard-coded"

test_gast_to_code.py
```diff
@@ -0,0 +1,169 @@
...
+
+gast_num = { "type": "num", "value": 47.47 }
+
+gast_true = {"type": "bool", "value": 1}
+
+gast_false = {"type": "bool", "value": 0}
+
+gast_arr = {
+	"type": "arr",
+	"elts":
```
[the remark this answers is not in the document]
jackdavidweber: elts is elements. We've all seen that as a common shorthand for elements, but we can add a TODO to change it if you think it's worth it

gast_to_code.py
jackdavidweber: got it!

gast_to_code.py
jackdavidweber: got rid of this when i cleaned up this function

gast_to_code.py
jackdavidweber: done!

PR #37, opened 2020-06-29
Adds error message if output language is not py or js

swalsh15:
https://tree.taiga.io/project/jackdavidweber-cjs-capstone/us/42?milestone=268292

### DIFF
--- a/main.py
+++ b/main.py
@@ -31,7 +31,9 @@ def main(input_code, input_lang, output_lang):
         #TODO: figure out hwo to do error messages
         return "Error must specify input language. For example, js for javascript and py for python"
 
-    if output_lang != "js" and output_lang != "py":
+    output_langs = ["js", "py"]
+    if output_lang not in output_langs:
+        # TODO: send 400 client error
         return "Error must specify output language. For example, js for javascript and py for python"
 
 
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -19,20 +19,20 @@
...
 
-    #TODO: also do error checking on the output language
     if input_lang == "js":
         gast = js_main.js_to_gast(input_code)
     elif input_lang == "py":
         gast = py_main.py_to_gast(input_code)
     else:
         #TODO: figure out hwo to do error messages
-        return "Error must specify language. For example, js for javascript and py for python"
+        return "Error must specify input language. For example, js for javascript and py for python"
```
Danish21, line 32: Consider returning a wrapper json object with two fields errorVal and successVal instead.  This way the client can easily tell if there is an error by checking if errorVal exists.

main.py
```diff
@@ -19,20 +19,20 @@
...
+
+    if output_lang != "js" and output_lang != "py":
```
g-harel, line 34: I feel like it would be easier to have an array of supported language symbols `["js", "py"]` and doing an `includes` check for the `input_lang` and `output_lang` at the start of the method.

main.py
```diff
@@ -19,20 +19,20 @@
...
+        return "Error must specify output language. For example, js for javascript and py for python"
```
g-harel, line 35: I think it would be awesome to also return an error code that indicates there was a client error (maybe 400).

PR #39, opened 2020-06-29
None/null functionality and test cases 

cjoshea9:


gast_to_code.py
```diff
@@ -84,6 +84,8 @@ def gast_router(gast, out_lang):
         return '"' + gast["value"] + '"'
     elif gast["type"] == "bool":
         return out["bool"][out_lang](gast)
+    elif gast["type"] == "none":
+        return "None" if out_lang == "py" else "null"
```
g-harel, line 88: Should probably move this out of here to keep up the pattern.

g-harel, line 88: Are we making a distinction between null and undefined in JS?

Danish21: +1. If not, I would add a comment stating that we are arbitrarily choosing null.

cjoshea9: We have decided to not worry about undefined for now and could possibly add in later. I will add a comment.

PR #41, opened 2020-06-29
Conditionals to gast

jackdavidweber:
added functionality for converting conditionals to gast. Still need to do gast_to_code for conditionals.


test_conditionals.py
```diff
@@ -0,0 +1,24 @@
...
+        py_code = 'if (1):\n\tprint("1 is true")\nelse:\n\tprint("1 is NOT true")\n'
+        self.assertEqual(py_code, main.main(js_code, 'js', 'py'))
+        self.assertEqual(js_code, main.main(py_code, 'py', 'js'))
+   
+    def test_elif(self):
+        js_code = 'if (1) {\n\tconsole.log("1 is true")\n} else if (2) {\n\tconsole.log("2 is true")\n\tconsole.log("second line")\n}'
+        py_code = 'if (1):\n\tprint("1 is true")\nelif (2):\n\tprint("2 is true")\n\tprint("second line")\n'
+        self.assertEqual(py_code, main.main(js_code, 'js', 'py'))
+        self.assertEqual(js_code, main.main(py_code, 'py', 'js'))
+
```
augerm, line 22: Does a combination of `if/else if/else` work? If so mind adding a test case for that.

translate/conditional/py_conditional.py
```diff
@@ -0,0 +1,10 @@
+import py_router as pr
+
+def if_statement(node):
+    print("you hit if)")
```
augerm: Remove print

test_conditionals.py
jackdavidweber: Yes it does- just added in a test!

PR #45, opened 2020-06-30
Add unsupported type to GAST. GAST to code prints unsupported message

swalsh15:
GAST adds node with type:error if feature is not supported. In code to gast output code is filled with error message.

### DIFF
--- a/gast_to_code.py
+++ b/gast_to_code.py
@@ -195,6 +195,11 @@ def gast_router(gast, out_lang):
         return binOp_helper(gast, out_lang)
     elif gast["type"] == "boolOp":
         return out["boolOp"][out_lang](gast)
+    elif gast["type"] == "error":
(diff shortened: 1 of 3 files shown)


gast_to_code.py
```diff
@@ -195,6 +195,11 @@ def gast_router(gast, out_lang):
         return binOp_helper(gast, out_lang)
     elif gast["type"] == "boolOp":
         return out["boolOp"][out_lang](gast)
+    elif gast["type"] == "error":
+        if gast["value"] == "unsupported":
+            # Error string
+            return "Feature not supported"
```
augerm, line 201: Can we make these error messages more explicit. Specifically it would be good to differentiate between the two messages.

"A language feature in your source code is unsupported by our translation service."
vs
"The source code fails to compile - please check your syntax."

PR #47, opened 2020-06-30
Not boolean logic

cjoshea9:
Add python to gast, js to gast, and gast to code functionality for the not / ! operator.


code_to_gast/helpers/py_helpers.py
```diff
@@ -61,7 +61,8 @@ def array(node):
 def pyop_to_str(op):
     m = {
             ast.Add: "+", ast.Mult: "*", ast.Div: "/", ast.Sub: "-",
-            ast.BitAnd: "&", ast.BitOr: "|", ast.And: "&&", ast.Or: "||"
+            ast.BitAnd: "&", ast.BitOr: "|", ast.And: "&&", ast.Or: "||",
```
g-harel, line 64: I know it's not from this CL, but the formatting of this makes it a bit more confusing than it needs to be. IMO, I would rename `m` to something more descriptive and give each mapping its own line.

cjoshea9: You are right, I will fix that.

PR #48, opened 2020-06-30
Refactor test files into one

cjoshea9:
Made two test files into one larger one to make sure test files aren't language specific and test both ways.


test/test_code_to_gast_conditionals.py
```diff
@@ -1,19 +1,20 @@
...
 
+import py_main
 import js_main
 
 
-class test_js_to_gast_conditionals(unittest2.TestCase):
+class test_code_to_gast_conditionals(unittest2.TestCase):
     maxDiff = None
-    def test_ifs(self):
+    def test_ifs_js(self):
```
jackdavidweber, line 17: maybe should each test for js and py be combined into single ones? Meaning just have test_ifs and then include both the python and javascript in the single test? That way each expected_gast does not have to be repeated

augerm:
Is there a reason they have to be in the same file? You could force yourself to write both tests even if they're in different files :)

My only concern is that when this file gets too large we're going to want to split it up anyways. But I don't have a strong opinion here so your call ha.

test/test_code_to_gast_conditionals.py
cjoshea9: Yeah that is a good idea. I did that in my other file but didn't think about it here.

PR #49, opened 2020-06-30
Rename functions with more description

cjoshea9:
Renamed functions to be more descriptive from code to gast and refactored boolOp and binOp functions in JS to be much simpler using power of recursion

### DIFF
--- a/code_to_gast/helpers/js_helpers.py
+++ b/code_to_gast/helpers/js_helpers.py
@@ -3,7 +3,7 @@ import js_router
 """
 handles arrays and recursively calls node_to_gast on all its elements
 """
-def js_array_expression(node):
+def js_array_expression_to_gast(node):
     gast = {"type" : "arr"}
     gast["elts"] = []
     for elm in node.elements:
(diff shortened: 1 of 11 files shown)


code_to_gast/helpers/js_helpers.py
```diff
@@ -3,7 +3,7 @@
 """
 handles arrays and recursively calls node_to_gast on all its elements
 """
-def js_array_expression(node):
+def js_array_expression_to_gast(node):
```
augerm, line 6: While you're here... I'd remove the `js_` prefix from these functions. 

Definitely counter to what I usually suggest but I think all call sites should look like `js_helpers.array_expression_to_gast` so the language won't be ambiguous.

code_to_gast/helpers/js_helpers.py
```diff
@@ -25,41 +25,23 @@ def program_to_gast(node):
 """
 converts a python ast BinOp and converts it to a gast node
 """
-def binOp(bop):
+def binOp_to_gast(bop):
```
augerm, line 28: `bop` --> `binOp_node` or probably better to go with `bin_op_node` all snake case while you're here.

gast_to_code/gast_to_code_main.py
```diff
@@ -8,79 +8,87 @@
...
+    return gast_to_node_router(gast["value"], "js") + "." + gast["id"] 
+
+def gast_to_py_logStatement(gast):
+    arg_string = gast_to_node_router(gast["args"],"py")
+    return "print(" + arg_string + ")"
 
-# js_specific_helpers
-def js_logStatement(gast):
-    arg_string = gast_router(gast["args"],"js")
+def gast_to_js_logStatement(gast):
```
augerm, line 47: `gast_to_js_log_statement`... in general the snake casing vs camel casing is a bit inconsistent.

You could save that for another PR though.

gast_to_code/gast_to_code_main.py
```diff
@@ -168,15 +167,15 @@ def js_unaryOp(gast):
 javascript: js
 python: py
 """
-def gast_router(gast, out_lang):
+def gast_to_node_router(gast, out_lang):
```
augerm, line 170: Could this be `gast_to_code` or `convert_gast_to_code`?

Danish21: +1. Router is an implementation detail which the caller should not care about.

code_to_gast/helpers/js_helpers.py
cjoshea9: Sounds good, just update that. We had some inconsistencies because we used camel case in the JSON object and then snake everywhere else

code_to_gast/helpers/js_helpers.py
cjoshea9: Resolved.

gast_to_code/gast_to_code_main.py
cjoshea9: Updated!

PR #51, opened 2020-07-01
Testing boolean and binary operations

swalsh15:
Added tests for boolean and bin ops. 
Fixed bug by nested the router in a try catch when building AST.
Redid logic for boolean op in python since the AST is not always nested.


main.py
```diff
@@ -36,7 +36,7 @@ def main(input_code, input_lang, output_lang):
         # TODO: send 400 client error
         return "Error must specify output language. For example, js for javascript and py for python"
 
-
+    print(gast)
```
cjoshea9: Remove print statement

test/test_bin_bool_ops.py
```diff
@@ -0,0 +1,54 @@
+import sys
+sys.path.insert(1, '/home/stephwalsh/capstone/cjs_capstone/')
+
+import unittest2
+import main
+
+class TestConditionals(unittest2.TestCase):
```
cjoshea9, line 7: Rename TestBinBoolOp

PR #54, opened 2020-07-06
Aug assign functionality 

cjoshea9:
Adds py to gast, js to gast, gast to code, and tests

### DIFF
--- a/code_to_gast/assign/py_assign.py
+++ b/code_to_gast/assign/py_assign.py
@@ -1,5 +1,5 @@
 import py_router as pr
-import py_helpers as helper
+import ast
 
 """
 takes python ast assigns and converts them to generic ast format
@@ -27,8 +27,20 @@ example:
 def aug_assign_to_gast(node):
     gast = {"type": "augAssign"}
     gast["left"] = pr.node_to_gast(node.target)
-    gast["op"] = helper.pyop_to_str(node.op) + "="
+    gast["op"] = augop_to_str(node.op)
     gast["right"] = pr.node_to_gast(node.value)
     return gast
+
+"""
+convert augmented assignment classes to strings
+"""
+def augop_to_str(op):
+    op_to_str_map = {
+        ast.Add: "+=", 
+        ast.Mult: "*=", 
+        ast.Div: "/=", 
+        ast.Sub: "-=",
+    }
+    return op_to_str_map[type(op)]
 
 
--- a/code_to_gast/assign/py_assign.py
+++ b/code_to_gast/assign/py_assign.py
@@ -1,4 +1,5 @@
 import py_router as pr
+import py_helpers as helper
 
 """
 takes python ast assigns and converts them to generic ast format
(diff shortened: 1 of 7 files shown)


code_to_gast/assign/py_assign.py
```diff
@@ -1,4 +1,5 @@
 import py_router as pr
+import py_helpers as helper
```
jackdavidweber, line 2: Not sure how I feel about importing py_helpers into py_assign. I thought the whole point was to have py_router handle all of the communication between files. It might start to get really confusing if files are calling functions in other files rather than only interacting with themselves and the py_router

jackdavidweber: I get why you want to do it since you are just using py_op_to_str. But maybe there is a way to put py_op_to_str in the router so that aug_assign_to_gast only needs to call the router to get access to it?

cjoshea9: I agree that it felt weird but I do not know the best solution since we need that helper across files. Maybe I just make another helper for the += -= *= and /= in the assign

test/test_var_assign.py
```diff
@@ -25,6 +25,10 @@ def test_boolean_assign(self):
     def test_none_assign(self):
         self.assertEqual('no = None', main.main('let no =null', 'js', 'py'))
         self.assertEqual('let help = null', main.main('help = None', 'py', 'js'))
+    
+    def test_aug_assign(self):
```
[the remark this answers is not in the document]
cjoshea9: Will do

code_to_gast/routers/js_router.py
```diff
@@ -41,6 +41,8 @@ def node_to_gast(node):
     #statements
     elif node.type == "VariableDeclaration":
         return js_assign.assign_to_gast(node)
+    elif node.type == "AssignmentExpression":
+        return js_assign.aug_assign_to_gast(node)
```
augerm, line 45: It looks like ["AssignmentExpressions"](https://esprima.readthedocs.io/en/4.0/syntax-tree-format.html?highlight=AssignmentExpression#assignment-expression) include regular `=` signs is that going to interfere with statements like `const x = 5`?

And if not... maybe `aug_assign_to_gast` should be renamed to `assignment_expression_to_gast`

cjoshea9: Statements like the one you mentioned fall under variable declaration in JS. A reassignment like simply x = 5 would be a AssignmentExpression and that would be translated into Python correctly because python treats all variable assignments the same.

test/test_var_assign.py
```diff
@@ -26,9 +26,13 @@ def test_none_assign(self):
         self.assertEqual('no = None', main.main('let no =null', 'js', 'py'))
         self.assertEqual('let help = null', main.main('help = None', 'py', 'js'))
     
-    def test_aug_assign(self):
+    def test_aug_assign_minus(self):
```
jackdavidweber, line 29: nice

code_to_gast/assign/py_assign.py
```diff
@@ -27,8 +27,20 @@ def assign_to_gast(node):
 def aug_assign_to_gast(node):
     gast = {"type": "augAssign"}
     gast["left"] = pr.node_to_gast(node.target)
-    gast["op"] = helper.pyop_to_str(node.op) + "="
+    gast["op"] = augop_to_str(node.op)
     gast["right"] = pr.node_to_gast(node.value)
     return gast
 
+"""
+convert augmented assignment classes to strings
```
jackdavidweber, line 35: yeah- i think this way makes a lot more sense

PR #55, opened 2020-07-06
Comparators

jackdavidweber:
We decided to represent comparators (i.e. 1 > 2) in the gast as binary operations, similar to how javascript does it. Due to the recursion, javascript to python was already working for comparators. However, since python implements comparators very differently, I had to do some work to convert the py ast to gast.

Also wrote test cases for comparators to check if things break in the future.

### DIFF
--- a/code_to_gast/helpers/py_helpers.py
+++ b/code_to_gast/helpers/py_helpers.py
@@ -162,6 +162,9 @@ def unary_op_to_gast(node):
 
 """
 takes node of type Compare and converts it to our generic ast representation
+example:
+    Example In: Compare(left=Num(n=1), ops=[Gt()], comparators=[Num(n=1), Num(n=2)])
+    Example Out: {'type': 'binOp', 'left': {'type': 'num', 'value': 1}, 'op': '>', 'right': {'type': 'num', 'value': 2}}
 """
 def compare_to_gast(node):
     # create list of comparators. (1>2>=3 would create list [1,2,3])
--- a/code_to_gast/helpers/py_helpers.py
+++ b/code_to_gast/helpers/py_helpers.py
@@ -173,6 +173,7 @@ def compare_to_gast(node):
 
     return compare_helper(comparator_list, op_list)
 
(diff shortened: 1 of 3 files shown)


code_to_gast/helpers/py_helpers.py
```diff
@@ -151,4 +158,34 @@ def name_to_gast(node):
...
+    # create list of comparators. (1>2>=3 would create list [1,2,3])
+    comparator_list = node.comparators
+    comparator_list.insert(0, node.left) # this is necessary since pyAST stores leftmost comparator seperately
+
+    # create list of operators. (1>2>=3 would create list [>, >=]
+    op_list = node.ops
+
+    return compare_helper(comparator_list, op_list)
+
+# TODO: combine logic in bool_op_helper and compare_helper into single function
```
g-harel, line 179: Is this for this PR or a future one?

test/test_bin_bool_ops.py
```diff
@@ -66,5 +67,30 @@ def test_combo_nested(self):
...
+        
+    def test_compare_equal(self):
+        js_code = '1 == 2'
+        py_code = '1 == 2'
+        self.assertEqual(py_code, main.main(js_code, 'js', 'py'))
+        self.assertEqual(js_code, main.main(py_code, 'py', 'js'))
+
+    def test_comparemulti_equal(self):
+        js_code = '1 == 2 == 3'
+        py_code = '1 == 2 == 3'
```
g-harel, line 91: Given this seems like the most delicate part of the python implementation, would be great to have a few more variations of more than one operators.

code_to_gast/helpers/py_helpers.py
jackdavidweber: for a future PR. Since Cory and Steph are also working on stuff at the same time, I've been trying to avoid going in and making edits to code that's unrelated to the feature I'm adding.

test/test_bin_bool_ops.py
jackdavidweber: Good idea- just did it!

PR #56, opened 2020-07-06
Added function declarations

swalsh15:
Added features for translating function declarations with arguments (no default parameters) for JS and Python. Planning to clean up files and add tests tm

### DIFF
--- a/code_to_gast/expression/py_expression.py
+++ b/code_to_gast/expression/py_expression.py
@@ -1,4 +1,5 @@
 import py_router as pr
+import astor
 
 """
 Takes python ast.expr node and converts them to the generic
@@ -30,4 +31,11 @@ handles attributes for python expressions
 def attribute_to_gast(node):
     gast = {"type": "attribute", "id": node.attr}
     gast["value"] = pr.node_to_gast(node.value)
+    return gast
+
+def function_def_to_gast(node):
+    gast = {"type": "functionDeclaration"}
+    gast["id"] = {"type": "name", "value": node.name}
+    gast["params"] = pr.node_to_gast(node.args)
+    gast["body"] = pr.node_to_gast(node.body)
     return gast
--- a/code_to_gast/expression/py_expression.py
+++ b/code_to_gast/expression/py_expression.py
@@ -1,5 +1,4 @@
 import py_router as pr
-import astor
 
 """
 Takes python ast.expr node and converts them to the generic
@@ -35,7 +34,7 @@ def attribute_to_gast(node):
 
 def function_def_to_gast(node):
     gast = {"type": "functionDeclaration"}
-    gast["id"] = {"type": "name", "value": node.name}
+    gast["id"] = pr.node_to_gast(node.name)
     gast["params"] = pr.node_to_gast(node.args)
     gast["body"] = pr.node_to_gast(node.body)
     return gast
--- a/code_to_gast/helpers/py_helpers.py
+++ b/code_to_gast/helpers/py_helpers.py
@@ -183,9 +183,8 @@ def arg_to_gast(node):
     return pr.node_to_gast(node.arg)
 
 """
-Takes native string class from python and creates gast node
-This node is used to store function names and arguments which
-is why the node type is name for identifier
+Python native string class is the type of function names and arguments. String literals are handled by str_to_gast()
+This function takes the name of an identifer and turns it into a gast node 
 """
 def str_to_gast(node):
     return {"type": "name", "value": node}
--- a/code_to_gast/helpers/py_helpers.py
+++ b/code_to_gast/helpers/py_helpers.py
@@ -51,7 +51,20 @@ def array_to_gast(node):
     for elem in ast.iter_child_nodes(node):
         if type(elem) != ast.Load:
             list_elem.append(pr.node_to_gast(elem))
-    gast["elts"] = list_elem
+    gast["elements"] = list_elem
+    return gast
+
+"""
+takes a dictionary converts it for the generic AST 
+"""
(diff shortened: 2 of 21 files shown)


code_to_gast/routers/py_router.py
```diff
@@ -51,4 +56,5 @@ def node_to_gast(node):
 
 
     else:
+        print(astor.dump(node))
```
augerm, line 59: Just making sure this is intentional however, I do think it would make sense to print out these error cases so we can pull them from logs later so I say keep it 👍

code_to_gast/expression/py_expression.py
```diff
@@ -1,4 +1,5 @@
 import py_router as pr
+import astor
```
g-harel, line 2: Is this being used?

code_to_gast/helpers/py_helpers.py
```diff
@@ -151,4 +151,10 @@ def name_to_gast(node):
 takes node of type unaryOp and converts it to our generic ast represenations
 """
 def unary_op_to_gast(node):
-    return {"type": "unaryOp", "op": pyop_to_str(node.op), "arg": pr.node_to_gast(node.operand)}
\ No newline at end of file
+    return {"type": "unaryOp", "op": pyop_to_str(node.op), "arg": pr.node_to_gast(node.operand)}
```
g-harel, line 154: Nit: Having newlines at the end of files will prevent this diff awkwardness where it looks like the last line is changed when it's just a newline.

code_to_gast/helpers/py_helpers.py
```diff
@@ -151,4 +151,10 @@ def name_to_gast(node):
...
+
+"""
+takes argument from function declaration and makes into gast node
+"""
+def arg_to_gast(node):
+    return {"type": "name", "value": node.arg}
```
g-harel, line 160: Does this handle complex values? seems like value should maybe come from another call to `node_to_gast`.

code_to_gast/js_main.py
```diff
@@ -20,3 +21,7 @@ def js_to_gast(js_input):
     except:
         return "Error: code could not compile"
     return js_router.node_to_gast(input_ast)
+
+#unformatted = (js_to_gast("function add(x, y){x + y}"))
```
g-harel, line 25: Nit: If this should stay, a TODO for clarification would be great.

code_to_gast/py_main.py
```diff
@@ -1,5 +1,6 @@
 import sys
 import ast
+import json
```
g-harel, line 3: Nit: Should this be removed?

gast_to_code/gast_to_code_main.py
```diff
@@ -111,6 +111,29 @@ def gast_to_js_if(gast):
 
     return out
 
+def gast_to_js_func_declarations(gast):
+    name = gast_to_code(gast["id"], "js")
+    args = gast_to_code(gast["params"], "js")
+    body = list_helper(gast["body"], "js", "\n\t")
+    out = "function " + name
+    out += "(" + args + ") {\n\t"
```
g-harel, line 119: This looks like it would only indent the first line of the body. I think formatting should be delegated to some other system to let this focus on generating correct code.

code_to_gast/helpers/py_helpers.py
cjoshea9: +1

gast_to_code/gast_to_code_main.py
swalsh15: Currently we only have support for one level of indent. The list_helper called in body separates each body statement with \n\t so every line is indented. We're added nested indents to the taiga and will address later.

code_to_gast/helpers/py_helpers.py
```diff
@@ -176,6 +176,47 @@ def name_to_gast(node):
...
+"""
+def arg_to_gast(node):
+    return pr.node_to_gast(node.arg)
+
+"""
+Takes native string class from python and creates gast node
+This node is used to store function names and arguments which
+is why the node type is name for identifier
+"""
+def str_to_gast(node):
```
jackdavidweber, line 190: This is slightly confusing to me since we already have ways of converting strings to gast. I don't think I'm quite understanding what you're doing here so maybe add some comments

code_to_gast/helpers/py_helpers.py
```diff
@@ -176,6 +176,47 @@ def name_to_gast(node):
...
+def return_statement_to_gast(node):
+    return {"type": "returnStatement", "value": pr.node_to_gast(node.value)}
+
+"""
+Handles args of function declarations with default values
+"""
+def function_args_to_gast(node):
+    return arg_helper(node.args, node.defaults, [])
+
+def arg_helper(arg_list, default_list, param_list):
```
jackdavidweber, line 205: These helpers (which we have all been using) go against the original principles of relying on the router. Wonder if there's a way to incorporate them into our router in some other PR.

test/test_function_declarations.py
```diff
@@ -0,0 +1,54 @@
...
+import unittest2
+import main
+
+class TestFunctionDeclarations(unittest2.TestCase):
+    def test_function_no_args(self):
+        js_code = 'function test() {\n\t2\n}'
+        py_code = 'def test():\n\t2'
+        self.assertEqual(py_code, main.main(js_code, 'js', 'py'))
+        self.assertEqual(js_code, main.main(py_code, 'py', 'js'))
+    def test_function_one_arg(self):
```
jackdavidweber, line 13: Feel like some spacing btwn functions would make this a lot more readable

PR #58, opened 2020-07-06
While loops

jackdavidweber:
Implemented while loop language feature using the gast specified in the [gast contract](https://docs.google.com/document/d/1Ycs8fz0tgYBZrnu2EKR8XvO3nq_6eW5jSDmBKLl37Mo/edit#heading=h.m5r5rietm0z)

Also wrote integration tests for going btwn python and javascript

### DIFF
--- a/code_to_gast/loop/js_loop.py
+++ b/code_to_gast/loop/js_loop.py
@@ -1,1 +1,17 @@
 import js_router as jr
+
+def while_statement_to_gast(node):
+    gast = {}
+    gast["type"] = "whileStatement"
+
+    """
+    Code re-used from js_conditional. TODO: write a helper or figure out how to go through router
+    """
+    body = node.body
+    if body.type == "BlockStatement":
+        gast["body"] = jr.node_to_gast(body)
+    else:
+        gast["body"] = [jr.node_to_gast(body)]
+
+    gast["test"] = jr.node_to_gast(node.test)
+    return gast
--- a/code_to_gast/loop/js_loop.py
+++ b/code_to_gast/loop/js_loop.py
@@ -1,0 +1,1 @@
+import js_router as jr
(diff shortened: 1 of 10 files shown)


code_to_gast/loop/js_loop.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    Code re-used from js_conditional. TODO: write a helper or figure out how to go through router
+    """
+    body = node.body
+    if body.type == "BlockStatement":
+        gast["body"] = jr.node_to_gast(body)
+    else:
+        gast["body"] = [jr.node_to_gast(body)]
+
+    gast["test"] = jr.node_to_gast(node.test)
```
Danish21, line 16: Is test the condition of the while loop? If so, IMO condition is a clearer name.

code_to_gast/routers/js_router.py
```diff
@@ -56,6 +58,9 @@ def node_to_gast(node):
     # Conditionals
     elif node.type == "IfStatement":
         return js_conditional.if_statement_to_gast(node)
+    # Loops
+    elif node.type == "WhileStatement":
```
Danish21, line 62: It would be nice to use enums here similar to the py router. Helps avoid typo bugs. For example, I noticed WhileStatement statement here has a capital W while js_loop.py uses lowercase w.

code_to_gast/loop/js_loop.py
jackdavidweber: Both python and javascript use "test" which is why we used it again. It's also what is used in if statements.

code_to_gast/routers/js_router.py
jackdavidweber: Yeah- I think the reason we did it that way in py_router is just because of the way the python ast library works. We would prob have to create a custom Enum which I've never done before but definitely could try to figure it out. Probably would be another PR though

jackdavidweber: Added to backlog!

PR #59, opened 2020-07-07
Array manipulation  - add and remove

cjoshea9:
Created built in attribute in the design doc and updated the functionality here

### DIFF
--- a/code_to_gast/expression/py_expression.py
+++ b/code_to_gast/expression/py_expression.py
@@ -33,10 +33,12 @@ def attribute_to_gast(node):
     if node.attr == "append":
         gast["type"] = "builtInAttribute"
         gast["id"] = "appendStatement"
+        return gast
     elif node.attr == "pop":
         gast["type"] = "builtInAttribute"
         gast["id"] = "popStatement"
+        return gast
     else:
         gast["type"] = "attribute"
         gast["id"] = node.attr
-    return gast
+        return gast
--- a/code_to_gast/expression/py_expression.py
+++ b/code_to_gast/expression/py_expression.py
@@ -34,11 +34,10 @@ def attribute_to_gast(node):
         gast["type"] = "builtInAttribute"
         gast["id"] = "appendStatement"
         return gast
-    elif node.attr == "pop":
+    if node.attr == "pop":
         gast["type"] = "builtInAttribute"
         gast["id"] = "popStatement"
         return gast
-    else:
-        gast["type"] = "attribute"
-        gast["id"] = node.attr
-        return gast
+    gast["type"] = "attribute"
+    gast["id"] = node.attr
+    return gast
--- a/code_to_gast/expression/py_expression.py
+++ b/code_to_gast/expression/py_expression.py
@@ -1,4 +1,5 @@
 import py_router as pr
+import py_built_in_functions as built_in
 
 """
 Takes python ast.expr node and converts them to the generic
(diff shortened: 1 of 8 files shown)


code_to_gast/expression/py_expression.py
```diff
@@ -28,6 +28,15 @@ def call_to_gast(node):
 handles attributes for python expressions
 """
 def attribute_to_gast(node):
-    gast = {"type": "attribute", "id": node.attr}
-    gast["value"] = pr.node_to_gast(node.value)
+    gast = {"value": pr.node_to_gast(node.value)} 
+
+    if node.attr == "append":
```
g-harel, line 33: Nit: I find it easier to reason about when functions return early instead of having "nested" conditionals.

```py
if node.attr == "append":
    ...
    return gast

if node.attr == "pop":
    ...
    return gast

...
```

cjoshea9: Okay I can update that.

code_to_gast/expression/py_expression.py
```diff
@@ -33,10 +33,12 @@ def attribute_to_gast(node):
     if node.attr == "append":
         gast["type"] = "builtInAttribute"
         gast["id"] = "appendStatement"
+        return gast
     elif node.attr == "pop":
```
g-harel, line 37: Still have the `else` blocks!

cjoshea9: Wait what do you mean?

g-harel: If you return in the `if` block, there is no need to use `elif`, you can just use another `if`. That's where the benefit is, since you don't have to keep the previous `if` context in mind.

cjoshea9: I see okay.

code_to_gast/helpers/js_helpers.py
```diff
@@ -73,11 +73,21 @@ def bool_op_to_gast(node):
...
 def member_expression_to_gast(node):
-  if node.property.name == "log":
+  if node.object.name == "console" and node.property.name == "log":
     return {"type": "logStatement"}
-
-  gast = {"type": "attribute", "id": node.property.name}
-  gast["value"] = js_router.node_to_gast(node.object)
+  
+  gast = {"value": js_router.node_to_gast(node.object)}
+  func_name = node.property.name
```
augerm, line 80: Adding a comment here explaining that this will mistranslate function calls that aren't Array.prototype.pop etc... might be useful down the line. Also consider adding a TODO with a link to a taiga ticket.

code_to_gast/helpers/js_helpers.py
```diff
@@ -73,11 +73,21 @@ def bool_op_to_gast(node):
 Used for functions called on objects and std funcs like console.log
 """
 def member_expression_to_gast(node):
-  if node.property.name == "log":
+  if node.object.name == "console" and node.property.name == "log":
```
augerm, line 76: Can `console.log` follow a similar approach? If it's overly complex then I'd just create a taiga ticket to standardize this.

code_to_gast/expression/py_expression.py
```diff
@@ -28,8 +28,18 @@ def call_to_gast(node):
 handles attributes for python expressions
 """
 def attribute_to_gast(node):
-    gast = {"type": "attribute", "id": node.attr}
-    gast["value"] = pr.node_to_gast(node.value)
+    gast = {"value": pr.node_to_gast(node.value)} 
+
+    if node.attr == "append":
```
augerm, line 33: Same comments here as in the JS file.

code_to_gast/helpers/js_helpers.py
cjoshea9: Updated

code_to_gast/helpers/js_helpers.py
cjoshea9: Since console.log we structured slightly differently then the other standard function calls, it looks slightly different but I will add a taiga ticket

code_to_gast/expression/py_expression.py
```diff
@@ -28,8 +28,18 @@ def call_to_gast(node):
...
+        gast["type"] = "builtInAttribute"
+        gast["id"] = "appendStatement"
```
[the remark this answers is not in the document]
cjoshea9: Should the gast["type"] assignment be an enum or something else? I am a little confused @Danish21

PR #60, opened 2020-07-08
For loops

jackdavidweber:
Full support (gast to code and code to gast) for forRange and forOf statements in javascript and python.

### For Range:
```
// javascript
for (let i=0; i<10; i+=1)

// python
for i in range(0,10,1):
```

### For Of:
```
// javascript
for (element of list)

// python
for element in list:
```

### DIFF
--- a/code_to_gast/loop/py_loop.py
+++ b/code_to_gast/loop/py_loop.py
@@ -28,67 +28,59 @@ def for_of_statement_to_gast(node):
     gast["body"] = pr.node_to_gast(node.body)
     return gast
 
+"""
+Converts for statement of the structure for i in range() to gast
+Notes about functionality
+    - only works with 2-3 range args
+    - only works if arg nodes are of type ast.Num (in case of positives) or 
+      ast.UnaryOp (in case of negatives). Does not work with vars.
+"""
 def for_range_statement_to_gast(node):
     gast = {}
-    gast["type"] = "forRangeStatement"
+    try:
+        gast["type"] = "forRangeStatement"
+        ast.dump(node)
+        args = node.iter.args
+        step_num = get_step_num(args)
+        start_num = arg_node_to_num(args[0])
+        end_num = arg_node_to_num(args[1])
 
-    args = node.iter.args
-    if len(args) < 2 or len(args) > 3:
-        return "unsupported: too many args in loop" # TODO: streamline messages
-    elif len(args) == 3:  
-        step_node = args[2]
-        
-        """
-        python stores step as ast.UnaryOp for negative numbers and
-        ast.Num for positive numbers. This logic takes care of this.
-        TODO: figure out a less "hacky" way of solving this problem
-        """
-        if type(step_node) == ast.UnaryOp:
-            step_num = -1 * step_node.operand.n
-        else:
-            step_num = step_node.n
+        gast["init"] = for_range_statement_init_helper(node.target, start_num)
+        gast["test"] = for_range_statement_test_helper(node.target, start_num, end_num)
+        gast["update"] = for_range_statement_update_helper(node.target, step_num)
+        gast["body"] = pr.node_to_gast(node.body)
 
-    # if there are only 2 arguments
-    else:
-        step_num = 1
+    except:
+        gast = {"type": "error", "value": "unsupported"}
 
-    # store the start and end number (in py ast node form) of the loop
-    start_node = args[0]
-    end_node = args[1]
-
-    gast["init"] = for_range_statement_init_helper(node.target, start_node)
-    gast["test"] = for_range_statement_test_helper(node.target, start_node, end_node)
-    gast["update"] = for_range_statement_update_helper(node.target, step_num)
-    gast["body"] = pr.node_to_gast(node.body)
-    return (gast)
+    return gast
 
 """
 Takes a node representing start of loop and the variable node.
 returns a gast node representing value for the init key in forRangeStatement gast
 Example: ast.parse("for i in range(0,10,2)")
-    input: var_node = ast.Name(i), start_node = ast.Num(0),
+    input: var_node = ast.Name(i), start_num = 10,
     output: {'type': 'varAssign', 'kind': 'let', 'varId': {'type': 'name', 'value': 'i'}, 'varValue': {'type': 'num', 'value': 0}}
 """
-def for_range_statement_init_helper(var_node, start_node):
+def for_range_statement_init_helper(var_node, start_num):
     gast = {}
     gast["type"] = "varAssign"
     gast["kind"] = "let"
     gast["varId"] = pr.node_to_gast(var_node)
-    gast["varValue"] = pr.node_to_gast(start_node)
+    gast["varValue"] = num_to_gast(start_num)
+
     return gast
 
 """
 Takes a variable node, start node and end node. Returns gast node
 representing value for the test key in forRangeStatement gast
 Example: ast.parse("for i in range(0,10,2)")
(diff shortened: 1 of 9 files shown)


test/test_code_to_gast_loops.py
```diff
@@ -0,0 +1,123 @@
...
+sys.path.append('code_to_gast/conditional')
+sys.path.append('code_to_gast/loop')
+sys.path.append('code_to_gast/helpers')
+sys.path.append('code_to_gast/routers')
+
+import py_main
+import js_main
+
+
+class test_code_to_gast_loops(unittest2.TestCase):
```
cjoshea9, line 15: Do you think we should continue testing, gast_to_code and code_to_gast or just test end to end?

test/test_code_to_gast_loops.py
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        self.assertEqual(expected_gast, py_main.py_to_gast(py_input))
+        self.assertEqual(expected_gast, js_main.js_to_gast(js_input))
+
+    def test_forOf(self):
+        js_input = 'for (elem of [1, 2]) {\n\t5\n}'
+        py_input = 'for elem in [1, 2]:\n\t5'
+        
+        expected_gast = {
```
cjoshea9, line 90: I like the way you are formatting these tests where you check back and forth between the same input for each language

code_to_gast/loop/py_loop.py
```diff
@@ -1,8 +1,91 @@
...
+    gast["body"] = pr.node_to_gast(node.body)
+    return gast
+
+
+def for_range_statement_to_gast(node):
+    gast = {}
+    gast["type"] = "forRangeStatement"
+
+    args = node.iter.args
+    if len(args) < 2 or len(args) > 3:
```
cjoshea9, line 36: Maybe add a comment here about why <2 and >3 is unsupported

code_to_gast/loop/py_loop.py
```diff
@@ -8,8 +8,13 @@ def while_statement_to_gast(node):
     gast["test"] = pr.node_to_gast(node.test)
     return gast
 
+"""
+Converts python for statements to gast
+Includes:
+    ForRangeStatements: for i in range(0,10,1)
+    ForOfStatements: for elem in [1,2,3]
+"""
```
cjoshea9, line 16: I like the documentation!

test/test_loops.py
```diff
@@ -17,6 +17,42 @@ def test_while_range(self):
         py_code = 'while (x < 10):\n\t5'
         self.assertEqual(py_code, main.main(js_code, 'js', 'py'))
         self.assertEqual(js_code, main.main(py_code, 'py', 'js'))
+        
+    def test_for_range_increment_one(self):
+        js_code = 'for (let i = 0; i < 10; i += 1) {\n\t5\n}'
```
augerm, line 22: Is i++ supported yet? 

If it its - Can we add a test?

If not - Can we add a Taiga ticket? I suspect this will be the most commonly typed for...loop so we might want to take that task on next.

code_to_gast/loop/py_loop.py
```diff
@@ -1,8 +1,128 @@
...
+    gast["iter"] = pr.node_to_gast(node.iter)
+    gast["body"] = pr.node_to_gast(node.body)
+    return gast
+
+def for_range_statement_to_gast(node):
+    gast = {}
+    gast["type"] = "forRangeStatement"
+
+    args = node.iter.args
+    if len(args) < 2 or len(args) > 3:
```
augerm, line 36: Can we invert this so this is the final else clause?

code_to_gast/loop/py_loop.py
```diff
@@ -1,8 +1,128 @@
...
+        return "unsupported: too many args in loop" # TODO: streamline messages
+    elif len(args) == 3:  
+        step_node = args[2]
+        
+        """
+        python stores step as ast.UnaryOp for negative numbers and
+        ast.Num for positive numbers. This logic takes care of this.
+        TODO: figure out a less "hacky" way of solving this problem
```
augerm, line 44: This seems reasonable to me - I'd say keep the comment remove the TODO but if you do keep the TODO I'd link to a taiga ticket

code_to_gast/loop/py_loop.py
augerm: +1 - would save us a little brain power :) 

Or even better move this check to a function with a descriptive name. Generally... I trust function names more than comments if you can sufficiently capture the reason in a function name.

gast_to_code/gast_to_code_main.py
```diff
@@ -187,6 +187,59 @@ def gast_to_js_while(gast):
...
+
+    # incrementor
+    incrementor_value = gast["update"]["right"]["value"]
+    incrementor_op = gast["update"]["op"] 
+    if incrementor_op == "-=":
+        incrementor = "-" + str(incrementor_value)
+    elif incrementor_op == "+=":
+        incrementor = str(incrementor_value)
+    else:
+        incrementor = "unsupported update operation" # TODO: rethink this error message
```
augerm, line 203: I'd rather handle this TODO now if you can.

code_to_gast/loop/py_loop.py
```diff
@@ -1,8 +1,128 @@
...
+        """
+        if type(step_node) == ast.UnaryOp:
+            step_num = -1 * step_node.operand.n
+        else:
+            step_num = step_node.n
```
Danish21, line 49: I believe that step_num could be anything right? It does not have to be a numeric literal....

For example,
ForRangeStatements: 
- for i in range(0,10,b)
- ForRangeStatements: for i in range(0,10,getStepNum())

Ok to not support that, but you should call it out in comments.

test/test_loops.py
cjoshea9: It is not supported yet, but we are planning to add it

test/test_code_to_gast_loops.py
jackdavidweber: I think we should continue. @Danish21 and I discussed on our 1:1 that when adding new languages it will make it a lot easier to have these tests as you just need to test from code to gast and back. Also, I think it is a good thing to get into the practice of doing and it helps my workflow even though it is time-consuming.

code_to_gast/loop/py_loop.py
jackdavidweber: Yes- I'll also pull it into separate function as per https://github.com/jackdavidweber/cjs_capstone/pull/60#discussion_r451866723

jackdavidweber:
Addressed comments and realized negative numbers weren't working so I addressed this and added test cases.

code_to_gast/loop/py_loop.py
```diff
@@ -1,8 +1,128 @@
...
+Takes a variable node, start node and end node. Returns gast node
+representing value for the test key in forRangeStatement gast
+Example: ast.parse("for i in range(0,10,2)")
+    input: var_node = ast.Name(i), start_node = ast.Num(0), end_node = ast.Num(10)
+    output: {'type': 'binOp', 'left': {'type': 'name', 'value': 'i'}, 'op': '<', 'right': {'type': 'num', 'value': 10}}
+"""
+def for_range_statement_test_helper(var_node, start_node, end_node):
+    # first need to figure out whether to use "<" or ">"
+    start_val = start_node.n
+    end_val = end_node.n
```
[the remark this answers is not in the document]
jackdavidweber: Got it- also I added in try catch to make sure it doesn't cause issues

PR #61, opened 2020-07-08
Array manipulation access

cjoshea9:
Adds look up and assignment for arrays. Need to decide if when translating from python to js if `arr[x] = 1` should be translated to `arr[x] = 1` or `let arr[x] = 1` .

### DIFF
--- a/code_to_gast/expression/py_expression.py
+++ b/code_to_gast/expression/py_expression.py
@@ -48,3 +48,6 @@ def function_def_to_gast(node):
     gast["params"] = pr.node_to_gast(node.args)
     gast["body"] = pr.node_to_gast(node.body)
     return gast
+
+def subscript_to_gast(node):
+    return {"type": "subscript", "index": pr.node_to_gast(node.slice.value)} # TODO expand to cover slices of a list
--- a/code_to_gast/expression/py_expression.py
+++ b/code_to_gast/expression/py_expression.py
@@ -50,4 +50,4 @@ def function_def_to_gast(node):
     return gast
 
 def subscript_to_gast(node):
-    return {"type": "subscript", "index": pr.node_to_gast(node.slice.value)} # TODO expand to cover slices of a list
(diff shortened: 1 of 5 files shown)


g-harel:
`let arr[x] = 1` would be pretty confusing, how hard would it be to remove that behavior?

cjoshea9:
It's just a simple if statement, I just added it

test/test_built_in_functions.py
```diff
@@ -13,6 +13,15 @@ def test_pop_statement(self):
         self.assertEqual('arr.pop(1)', main.main('arr.pop(1)', 'py', 'js'))
         self.assertEqual('mine.pop(True)', main.main('mine.pop(true)', 'js', 'py'))
     
+    def test_array_look_up(self):
+        self.assertEqual('arr[x]', main.main('arr[x]', 'js', 'py'))
+        self.assertEqual('array[3]', main.main('array[3]', 'py', 'js'))
+    
+    def test_array_assign(self):
+        self.assertEqual('arr[x] = 6', main.main('arr[x] = 6', 'js', 'py'))
+        #TODO since we don't have context it adds the let, do we want to change that?
```
augerm, line 22: This comment can probably be removed now.

code_to_gast/expression/py_expression.py
```diff
@@ -48,4 +48,7 @@ def function_def_to_gast(node):
     gast["id"] = pr.node_to_gast(node.name)
     gast["params"] = pr.node_to_gast(node.args)
     gast["body"] = pr.node_to_gast(node.body)
-    return gast
\ No newline at end of file
+    return gast
+
+def subscript_to_gast(node):
+    return {"type": "subscript", "index": pr.node_to_gast(node.slice.value), "value": pr.node_to_gast(node.value)} # TODO expand to cover slices of a list
```
augerm, line 54: Can we move the TODO above the return statement?

augerm: Also might be easier to read if we make the dictionary multiline 
```
{
    "type": "subscript",
    ... etc
}
```

cjoshea9: Will do.

PR #65, opened 2020-07-09
Add more built in functions 

cjoshea9:
Add string manipulation and array manipulation functions to the enum and test cases

### DIFF
--- a/js_built_in_functions.py
+++ b/js_built_in_functions.py
@@ -5,7 +5,16 @@ All JavaScript built in functions we support mapped to their corresponding
 generic AST string
 """
 class js_built_in_functions(enum.Enum):
+    #Array manipulation
     push = "appendStatement"
     pop = "popStatement"
     sort = "sortStatement"
     concat = "extendStatement"
+    reverse = "reverseStatement"
+
+    #String manipulation
+    search = "findStatement"
+    split = "splitStatement"
+    toUpperCase = "upperStatement"
+    toLowerCase = "lowerStatement"
+    indexOf = "indexStatement"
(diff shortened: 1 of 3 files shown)


js_built_in_functions.py
```diff
@@ -5,5 +5,16 @@
 generic AST string
 """
 class js_built_in_functions(enum.Enum):
+    #Array manipulation
     push = "appendStatement"
```
Danish21, line 9: Uber nit: Is it possible for the values of this enum to also be enums to avoid typo mistakes between the .js and .py file?

test/test_built_in_functions.py
```diff
@@ -21,6 +21,38 @@ def test_array_assign(self):
         self.assertEqual('arr[x] = 6', main.main('arr[x] = 6', 'js', 'py'))
         self.assertEqual('array[3] = []', main.main('array[3] = []', 'py', 'js')) 
     
+    def test_sort(self):
+        self.assertEqual('myArr.sort()', main.main('myArr.sort()', 'js', 'py'))
```
g-harel, line 25: What if the sort method is used on a non-array, like for example a custom class that happens to also have that method?

augerm: I think we get burned in that case. But we discussed this and the plan is to punt on that for now, it's a pretty big/ambiguous task and there may be lower hanging fruit.

With that said @cjoshea9 can we create a ticket for this? As we add more types users are more likely to run into these cases.

cjoshea9: Yes, we decided to punt on that for now, but I will make a ticket to address this later. @g-harel @augerm

js_built_in_functions.py
cjoshea9: Yes, but I thought that would make it less readable. Do you think I should still do it? @Danish21

PR #66, opened 2020-07-10
Collect user feedback

swalsh15:
Fixed requirements and imports so firebase works when deployed on heroku. Added changes so firebase only is initiated once and entries that don't compile or are empty ie "" are not stored.

### DIFF
--- a/data_collection.py
+++ b/data_collection.py
@@ -1,4 +1,4 @@
-'''import firebase_admin
+import firebase_admin
 from firebase_admin import credentials
 from firebase_admin import firestore
 import json
@@ -37,4 +37,3 @@ def store_query(input_code, output_code, input_lang, output_lang):
         u'output_lang': output_lang,
         u'successful_translate': successful_translate
     })
-'''
--- a/data_collection.py
+++ b/data_collection.py
@@ -13,8 +13,10 @@ def store_query(input_code, output_code, input_lang, output_lang):
         print("Please set GOOGLE_APPLICATION_CREDENTIALS environmental variable. See readme for more info")
         return
     
-    cred = credentials.Certificate(cred_object)
-    firebase_admin.initialize_app(cred)
+    # only initialize firebase once
+    if not firebase_admin._apps:
+        cred = credentials.Certificate(cred_object)
+        firebase_admin.initialize_app(cred)
 
     # only store code written on live heroku
     if "SHOULD_WRITE_TO_DATABASE" not in os.environ:
--- a/data_collection.py
+++ b/data_collection.py
@@ -20,7 +20,6 @@ def store_query(input_code, output_code, input_lang, output_lang):
 
     # only store code written on live heroku
     if "SHOULD_WRITE_TO_DATABASE" not in os.environ:
-        print("locall dev")
         return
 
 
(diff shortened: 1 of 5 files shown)


main.py
```diff
@@ -38,14 +38,19 @@ def main(input_code, input_lang, output_lang):
...
-    # don't return error without sending to database
-    output_code = ""
+
     if (type(gast) == str) :
-        # return error if gast not built
-        output_code = "Error: did not compile"
-    else:
-        output_code = gtc.gast_to_code(gast, output_lang)
+        # return error if gast not built - dont store in database
+        return "Error: did not compile"
```
g-harel, line 44: I'd prefer if there was a stronger link between this error response and the condition being checked in the data_collection code. Maybe a constant string or error code?

data_collection.py
```diff
@@ -13,8 +13,10 @@ def store_query(input_code, output_code, input_lang, output_lang):
         print("Please set GOOGLE_APPLICATION_CREDENTIALS environmental variable. See readme for more info")
         return
     
-    cred = credentials.Certificate(cred_object)
-    firebase_admin.initialize_app(cred)
+    # only initialize firebase once
+    if not firebase_admin._apps:
```
augerm, line 17: I would wrap this function in a `DataService` [singleton](https://python-3-patterns-idioms-test.readthedocs.io/en/latest/Singleton.html) class. That way it is only initialized once. Happy to hop on a call and walk through this!

tools/add_environ_var.py
```diff
@@ -1,11 +1,17 @@
...
+import subprocess
 
 try:
     with open("../private_keys/firebase-access-key.json") as json_file:
         data = json.load(json_file)
+        env = {
+            **os.environ,
+            "GOOGLE_APPLICATION_CREDENTIALS": json.dumps(data),
+        }
+        subprocess.Popen('/bin/bash', env=env).wait()
```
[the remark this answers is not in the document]
swalsh15: This exports the variable and allows it to be used in subprocesses. The old code created a local variable that could only be used in the file it was created in

main.py
```diff
@@ -38,14 +38,19 @@ def main(input_code, input_lang, output_lang):
...
    
-    # don't return error without sending to database
-    output_code = ""
+
     if (type(gast) == str) :
-        # return error if gast not built
-        output_code = "Error: did not compile"
-    else:
-        output_code = gtc.gast_to_code(gast, output_lang)
+        # return error if gast not built - dont store in database
```
jackdavidweber, line 43: why not store this in the database? Wouldn't it be a useful thing to know if people aren't able to get code to compile?

main.py
```diff
@@ -38,14 +38,19 @@ def main(input_code, input_lang, output_lang):
...
+        return "Error: did not compile"
+   
+    output_code = gtc.gast_to_code(gast, output_lang)
     
-    #data_collection.store_query(input_code, output_code, input_lang, output_lang)
+    # don't store empty strings
```
jackdavidweber, line 48: Maybe elaborate on comment

main.py
swalsh15: Good idea - going to store and categorize whether successful, didn't compile, or feature not supported

main.py
swalsh15: Got it will add to it

PR #71, opened 2020-07-13
Router refactor

cjoshea9:
Refactor whole code base to separate code into python and javascript specific files and add the register pattern.


shared/gast_to_code/gast_to_code_router.py
```diff
@@ -0,0 +1,167 @@
...
+    elif gast["type"] == "assignPattern":
+        return converter.handle_assign_pattern(gast)
+    elif gast["type"] == "error":
+        if gast["value"] == "unsupported":
+            # Error string
+            return "Feature not supported"
+        return "Error"
+
+
+# out = {
```
jackdavidweber, line 86: We should remove this

cjoshea9: Updated.

app.py
```diff
@@ -14,6 +15,7 @@
 
 class Translate(Resource):
     def get(self):
+        bootstrap()
```
jackdavidweber, line 18: Although I was the one to suggest this, I now am thinking that maybe we should add a constructor so that bootstrap is called whenever Translate class is invoked - even if get request isn't. I just added a TODO in my current branch so do not change this now as it will create later merge issues, but wanted to just document this somewhere

cjoshea9: Sounds good, and I agree that seems like a better idea

jackdavidweber: Yup! I actually just fixed it on my current language detection branch so it will be resolved shortly!

test_main.py
```diff
@@ -1,4 +1,7 @@
 import unittest2
+from bootstrap import bootstrap
+
+bootstrap()
```
Danish21, line 4: Add a comment for posterity on why this is needed.

app.py
Danish21: +1 to this

test_main.py
cjoshea9: Fixed

shared/gast_to_code/converter_registry.py
```diff
@@ -0,0 +1,12 @@
+
+class ConverterRegistry():
```
g-harel, line 2: Could add `get_langs()` to show in the UI.

jackdavidweber: @g-harel already did this in language detection branch!

PR #73, opened 2020-07-14
Bash boilerplate

cjoshea9:


cjoshea9:
After looking at lots of examples in Bash, we may have to try and do another language. The AST treats things much differently than the javascript and python ASTs which is fine but would just require a lot of hard-coding.

bash/code_to_gast/bash_helpers.py
```diff
@@ -0,0 +1,17 @@
+
+def word_to_gast(node):
+    literal = node.word
+    if literal == 'true':
+        return {'type': 'bool', 'value': 1}
+    elif literal == 'false':
+        return {'type': 'bool', 'value': 0}
+    else:
+        try:
```
Danish21, line 9: Nit:
For better readability consider helper functions to avoid nested try and excepts. See https://stackoverflow.com/questions/1265665/how-can-i-check-if-a-string-represents-an-int-without-using-try-except

bash/code_to_gast/bash_router.py
```diff
@@ -0,0 +1,12 @@
+import bashlex
+
+def node_to_gast(node):
+
+    print (type(node))
```
Danish21: Remove?

cjoshea9: I removed that in one of the later commits

bash/code_to_gast/bash_helpers.py
cjoshea9: I had that before, but right now we are counting floats and ints the same way so I think I need the try and except.

cjoshea9: I could put it in a helper function though

PR #74, opened 2020-07-14
Boilerplate for java for primitives and print

swalsh15:
Boilerplate for adding java to translation. There is support for adding primitives - strings, booleans, numbers (ints) and System.out.println() statements with one argument. There are currently limitations with only one class and one function being supported. At the moment I'm just translating code inside the main function (on the live server we will either artificially wrap or add the class/function to the text box).

### DIFF
--- a/java/code_to_gast/java_router.py
+++ b/java/code_to_gast/java_router.py
@@ -1,0 +1,43 @@
+import javalang.tree
+import java.code_to_gast.java_expression as java_expression
+
+def node_to_gast(node):
+    #base cases
+    if type(node) == javalang.tree.Literal:
+        # node.value is stored as string in java AST
+        if node.value.isnumeric():
+            return {"type": "num", "value": node.value}
+        if node.value.startswith('"'):
+            return {"type": "str", "value": node.value.replace('"', '')}
+        elif node.value == "true":
+            return {"type": "bool", "value": 1}
+        elif node.value == "false":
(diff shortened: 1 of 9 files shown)


java/code_to_gast/java_router.py
```diff
@@ -0,0 +1,43 @@
...
+        if node.value.isnumeric():
+            return {"type": "num", "value": node.value}
+        if node.value.startswith('"'):
+            return {"type": "str", "value": node.value.replace('"', '')}
+        elif node.value == "true":
+            return {"type": "bool", "value": 1}
+        elif node.value == "false":
+            return {"type": "bool", "value": 0}
+        else:
+            return "Unsupported prim"
```
Danish21, line 18: Suggestion: s/prim/primitive to avoid any confusion.

java/gast_to_code/gast_to_code_java.py
```diff
@@ -0,0 +1,69 @@
...
+        return "System.out.println"
+
+    def handle_var_assign(gast):
+        pass
+
+    def handle_aug_assign(gast):
+        pass
+
+    def handle_func_call(gast):
+        return router.gast_to_code(gast["value"], "java") + "(" + router.gast_to_code(gast["args"], "java") + ")"
```
Danish21, line 36: Consider making this a constant at the top of the file since it is used in multiple places.

PR #77, opened 2020-07-15
Variable declarations for primitive vals

swalsh15:
Added support for variable declarations for strings, booleans, and ints.

### DIFF
--- a/java/code_to_gast/java_router.py
+++ b/java/code_to_gast/java_router.py
@@ -1,6 +1,7 @@
 import javalang.tree
 import java.code_to_gast.java_expression as java_expression
 import java.code_to_gast.java_helpers as java_helpers
+import java.code_to_gast.java_assign as java_assign
 
 def node_to_gast(node):
     #base cases
@@ -30,6 +31,11 @@ def node_to_gast(node):
         return node_to_gast(node.expression)
     elif type(node) == list:
         return java_helpers.node_list_to_gast_list(node)
+    elif type(node) == javalang.tree.LocalVariableDeclaration:
+        # our current gAST doesn't support multiple declarations/unsure what that is
+        return node_to_gast(node.declarators[0])
+    elif type(node) == javalang.tree.VariableDeclarator:
+        return java_assign.assign_to_gast(node)
     else:   
         # not supported
         return {"type": "error", "value": "unsupported"}
--- a/java/code_to_gast/java_router.py
+++ b/java/code_to_gast/java_router.py
@@ -32,8 +32,11 @@ def node_to_gast(node):
     elif type(node) == list:
         return java_helpers.node_list_to_gast_list(node)
     elif type(node) == javalang.tree.LocalVariableDeclaration:
-        # our current gAST doesn't support multiple declarations/unsure what that is
-        return node_to_gast(node.declarators[0])
+        # our current gAST doesn't support multiple declarations
+        if len(node.declarators) == 1:
+            return node_to_gast(node.declarators[0])
+        else:
+            return {"type": "error", "value": "unsupported"}
     elif type(node) == javalang.tree.VariableDeclarator:
         return java_assign.assign_to_gast(node)
     else:   
--- a/java/gast_to_code/gast_to_code_java.py
+++ b/java/gast_to_code/gast_to_code_java.py
@@ -27,7 +27,20 @@ class JavaGastToCodeConverter(AbstractGastToCodeConverter):
         return "System.out.println"
 
     def handle_var_assign(gast):
-        pass
+        var_id = router.gast_to_code(gast["varId"], "js")
+        var_value = router.gast_to_code(gast["varValue"], "js")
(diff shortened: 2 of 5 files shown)


java/code_to_gast/java_router.py
```diff
@@ -30,6 +31,11 @@ def node_to_gast(node):
         return node_to_gast(node.expression)
     elif type(node) == list:
         return java_helpers.node_list_to_gast_list(node)
+    elif type(node) == javalang.tree.LocalVariableDeclaration:
+        # our current gAST doesn't support multiple declarations/unsure what that is
```
g-harel, line 35: I don't have any fixes to suggest, but this seems like it might lead to silent failures.

java/gast_to_code/gast_to_code_java.py
```diff
@@ -27,7 +27,20 @@ def handle_log_statement(gast):
         return "System.out.println"
 
     def handle_var_assign(gast):
-        pass
+        var_id = router.gast_to_code(gast["varId"], "js")
```
g-harel, line 30: Is reusing the `js` helpers here intentional?

java/gast_to_code/gast_to_code_java.py
```diff
@@ -27,7 +27,20 @@ def handle_log_statement(gast):
...
+        var_value = router.gast_to_code(gast["varValue"], "js")
+        var_type = gast["varValue"]["type"]
+        
+        if var_type == "num":
+            kind = "int"
+        elif var_type == "str":
+            kind = "String"
+        elif var_type == "bool":
+            kind = "boolean"
+        else:
```
g-harel, line 40: Should this be an unsupported error instead? At least for the type part.

java/gast_to_code/gast_to_code_java.py
```diff
@@ -27,7 +27,20 @@ def handle_log_statement(gast):
...
+            kind = "customType"
+
+        return kind + " " + var_id + " = " + var_value
```
g-harel, line 43: Would this be the right place to add semicolons?

java/code_to_gast/java_router.py
```diff
@@ -30,6 +31,11 @@ def node_to_gast(node):
...
+        return node_to_gast(node.declarators[0])
```
jackdavidweber, line 36: Don't love the idea of having to call the first item of the list. Feel like when we have done this in the past, we later realized that the other items in the list were actually important and necessary. May be necessary in this case but just something to be mindful of

java/gast_to_code/gast_to_code_java.py
jackdavidweber: +1 'js' should be reserved for javascript

test/test_java_primitives.py
```diff
@@ -8,6 +8,8 @@ def test_basic(self):
     def test_basic_args(self):
         code = 'System.out.println("s");\n 1;'
         self.assertEqual('System.out.println("s")\n1', main.main(code, 'java', 'java'))
-
+    def test_basic_var_declaration(self):
+        code = 'int x = 1;\n String s = "s";\nboolean b = false;'
+        self.assertEqual('int x = 1\nString s = "s"\nboolean b = false', main.main(code, 'java', 'java'))
```
jackdavidweber, line 13: I would split this into multiple tests to improve readability and make it easier to fix future errors

java/code_to_gast/java_router.py
Danish21: https://stackoverflow.com/questions/6202818/initializing-multiple-variables-to-the-same-value-in-java#:~:text=You%20can%20declare%20multiple%20variables,opposite%20of%20%22visually%20simple%22.

Maybe detect that case and return feature not supported error?

java/gast_to_code/gast_to_code_java.py
Danish21: I think this is for when the kind is not a primitive i.e. MyFooClass. Can we store/use that exact type instead? i.e. 

java -> java would be
MyFooClass foo = new FooClass(); -> MyFooClass foo = new FooClass(); 
instead of customType foo = new FooClass();

java/gast_to_code/gast_to_code_java.py
Danish21: +1 not sure how semicolons are working.

java/code_to_gast/java_router.py
Danish21: Maybe also file a ticket for js and python since they both support though with different syntaxs.

java/code_to_gast/java_router.py
swalsh15: I'll look more into this. BTW this is still being done on the python and js side where we only look at target[0] or declaration[0]. I'll leave the comment in

swalsh15: Adding an error check here to return unsupported if more than one elm in list

PR #79, opened 2020-07-15
Bash js boiler 

cjoshea9:
Changed to use javascript instead of python to do the bash code to gast and then called the javascript files using os in main.py


g-harel:
Please gitignore `node_modules`

cjoshea9:
I committed node_modules on accident, I should delete those right?

jackdavidweber:
Make sure to merge this into bash development branch rather than master!

bash/code_to_gast/bash_helpers.js
```diff
@@ -0,0 +1,46 @@
...
+    return node_list
+}
+
+function word_to_gast(node) {
+    text = node["text"]
+    if (text === "true") {
+        return {"type": "bool", "value": 1}
+    } else if (text === "false") {
+        return {"type": "bool", "value": 0}
+    } else if (!isNaN(text)) {
```
jackdavidweber, line 24: this seems dangerous... maybe add comment explaining what is happening here so that it is easier to fix if it breaks

bash/code_to_gast/bash_helpers.js
```diff
@@ -0,0 +1,46 @@
...
+    let node_list = []
+    for(const elem of node) {
+        node_list.push(bash_router.node_to_gast(elem))
+    }
+    return node_list
+}
+
+function word_to_gast(node) {
+    text = node["text"]
+    if (text === "true") {
```
jackdavidweber, line 20: Why is text being handled within this function as opposed to by the router? Could you just return code_to_gast(text) and then have a separate function rather than dealing with it within here?

bash/code_to_gast/bash_helpers.js
```diff
@@ -0,0 +1,46 @@
...
+    } else if (text === "false") {
+        return {"type": "bool", "value": 0}
+    } else if (!isNaN(text)) {
+        return {"type": "num", "value": Number(text)}
+    } else {
+        return {"type": "str", "value": text}
+    }
+}
+
+function command_to_gast(node) {
```
jackdavidweber, line 31: Good way of dealing with this issue! I think because bash is so weird, you unfortunately probably should add some more comments / examples / documentation.

bash/code_to_gast/bash_helpers.js
```diff
@@ -0,0 +1,46 @@
...
+        gast = {"type": "funcCall"}
+        gast["args"] = bash_router.node_to_gast(node["suffix"])
+        gast["value"] = {"type": "logStatement"}
+        return gast
+    } else {
+        return bash_router.node_to_gast(node["name"])
+    }
+}
+
+exports.script_to_gast = script_to_gast
```
jackdavidweber, line 43: should this be wrapped in a function?

bash/code_to_gast/bash_main.js
```diff
@@ -0,0 +1,12 @@
...
+
+const bash_input = process.argv[2]
+let input_ast = ""
+try {
+    input_ast = parse(bash_input);
+} catch(err) {
+    return "Error: code could not compile"
+}
+
+console.log(JSON.stringify(bash_router.node_to_gast(input_ast)))
```
jackdavidweber, line 12: remove? JSON.stringify should probably not exist within the console.log since it is an important call.

main.py
```diff
@@ -2,10 +2,11 @@
 import javascript.code_to_gast.js_main as js_main
 import python.code_to_gast.py_main as py_main
 import java.code_to_gast.java_main as java_main
-import bash.code_to_gast.bash_main as bash_main
 import shared.gast_to_code.gast_to_code_router as gtc
 from data_service import DataService
 from bootstrap import bootstrap
+import subprocess
+import json
```
jackdavidweber, line 9: do these imports need to be added to requirements.txt?

main.py
```diff
@@ -25,7 +26,10 @@ def main(input_code, input_lang, output_lang):
     elif input_lang == "java":
         gast = java_main.java_to_gast(input_code)
     elif input_lang == "bash":
-        gast = bash_main.bash_to_gast(input_code)
+        process = subprocess.Popen(['node', 'bash/code_to_gast/bash_main.js', input_code], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
jackdavidweber, line 29: this is fine for now, but feel like this should probably eventually live somewhere else

test/test_log_statement.py
```diff
@@ -9,6 +9,7 @@ def test_no_args(self):
     def test_one_arg(self):
         self.assertEqual('console.log("hello")', main.main('print("hello")', 'py', 'js'))
         self.assertEqual('print(5)', main.main('console.log(5)', 'js', 'py'))
+        self.assertEqual('echo 8', main.main('console.log(8)', 'js', 'bash'))
```
jackdavidweber, line 12: This is the problem with only doing these big integration tests. Need to now ask the question if we should translate btwn all the combinations of language pairs in order to test. Not sure there is a good answer and I think integration tests are important. @Danish21 we talked about this during our 1:1. Do you have any advice?

bash/code_to_gast/bash_helpers.js
cjoshea9: Okay

bash/code_to_gast/bash_helpers.js
cjoshea9: This is an attribute of the word node, and word nodes are the base cases, I will add a comment so it is more clear

bash/code_to_gast/bash_helpers.js
cjoshea9: No, it needs to be directly exported in the file so other files can see it. I could wrap them in a function then call the function right below it if you think that is cleaner

main.py
cjoshea9: No these are built in

test/test_log_statement.py
cjoshea9: I agree that our testing framework is not perfect. Also, Michael and I decided it may not be worth it to continue to have bash as an input language, so we will definitely need gast_to_code tests for bash

bash/code_to_gast/bash_helpers.js
```diff
@@ -0,0 +1,59 @@
+const bash_router = require("./bash_router.js")
+
+/**
+ * Handles the script type which is the top level node that every command has
+ */
+function script_to_gast(node) {
+    gast = {"type": "root"}
```
g-harel, line 7: Would prefer if `gast` was fully declared with `const`.

bash/code_to_gast/bash_helpers.js
g-harel: I had no idea `isNaN` has some fancy stuff when passed a string ...

bash/code_to_gast/bash_helpers.js
g-harel: If you wanted to do this inline, I think this would also do the trick:

```js
exports.dothething = function () {}
```

bash/code_to_gast/bash_main.js
```diff
@@ -0,0 +1,14 @@
+const bash_router = require("./bash_router.js")
+const parse = require("bash-parser");
+
+const bash_input = process.argv[2]
```
g-harel, line 4: Should this fail when given too little arguments?

bash/code_to_gast/bash_main.js
```diff
@@ -0,0 +1,14 @@
...
+let input_ast = ""
+try {
+    input_ast = parse(bash_input);
+} catch(err) {
+    return "Error: code could not compile"
+}
+
+const gast = bash_router.node_to_gast(input_ast)
+const json_string = JSON.stringify(gast)
+console.log(json_string)
```
g-harel, line 14: Is this how the result is sent back to the caller?

bash/gast_to_code/bash_gast_to_code_converter.py
```diff
@@ -10,4 +10,10 @@ def handle_bool(gast):
             return "true"
         else:
             return "false"
+    
+    def handle_log_statement(gast):
+        return "echo"
```
g-harel, line 15: Is this intentionally not printing the arguments?

main.py
g-harel, line 29: Would like this logic to be abstracted away into a function with the same signature as the other ones.

test/test_log_statement.py
g-harel: I think it would be valuable to also do some lang <-> gast tests if those don't exist already. 
Tests are pretty cheap so why not also have a bunch of lang -> lang tests as well.

bash/code_to_gast/bash_helpers.js
cjoshea9: Fixed in next PR

bash/code_to_gast/bash_main.js
cjoshea9: This will always be given three arguments, `['node', 'bash/code_to_gast/bash_main.js', input_code]`. The third argument is always the input_code

bash/code_to_gast/bash_main.js
cjoshea9: Yes

bash/gast_to_code/bash_gast_to_code_converter.py
cjoshea9: Yes, this is how we do it in every language. Arguments are handled by another recursive call to the router

main.py
cjoshea9: The problem is that this is python code calling javascript code, so I can't put the logic in bash_main because it is a js file. I could make a helper in main.py but that seems awkward

g-harel: Why would a helper be awkward? Doesn't need to be in `main.py`.

bash/code_to_gast/bash_main.js
g-harel: What if the integration is accidentally broken and it's given incorrect args? Solid validation could prevent silent errors.

test/test_log_statement.py
cjoshea9: We aren't going to do bash as an input language anymore but I think those are valuable for gast to code

PR #82, opened 2020-07-16
Java custom functions

swalsh15:
Add support for calling custom functions in java. Works for calling static and non static functions with arguments.

### DIFF
--- a/java/code_to_gast/java_expression.py
+++ b/java/code_to_gast/java_expression.py
@@ -5,6 +5,7 @@ Takes method invocation ie function declaration and translates to
 gAST node of type funcCall
 '''
 def method_invocation_to_gast(node):
+    
     gast = {"type": "funcCall"}
     gast["args"] = java_router.node_to_gast(node.arguments)
     
@@ -12,8 +13,27 @@ def method_invocation_to_gast(node):
     if node.qualifier == "System.out" and node.member == "println":
         gast["value"] = {"type": "logStatement"}
     else:
-        gast["value"] = {"type": "customStatement"}
+        # function not called on object
+        if not node.qualifier:
+            gast["value"] = {"type": "name", "value": node.member}
+        else:
+            object_list = (node.qualifier.split("."))
+            object_list.append(node.member)
+            gast["value"] = functionHelper(object_list)
     return gast
+
+def functionHelper(object_list):
+    gast = {}
+    gast["type"] = "attribute"
+    gast["id"] = object_list.pop()
+
+    if len(object_list) == 1:
+        gast["value"] = {"type": "name", "id": object_list.pop()}
+    else: 
+        gast["value"] = functionHelper(object_list)
+    
+    return gast
+
 
 '''
 Takes class declaration and turns into start of gast
--- a/java/code_to_gast/java_expression.py
+++ b/java/code_to_gast/java_expression.py
@@ -5,7 +5,6 @@ Takes method invocation ie function declaration and translates to
 gAST node of type funcCall
 '''
 def method_invocation_to_gast(node):
-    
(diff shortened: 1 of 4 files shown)


test/test_java_function.py
```diff
@@ -0,0 +1,26 @@
+import unittest2
+import main
+
+class TestJavaFunctionCalls(unittest2.TestCase):
```
jackdavidweber, line 4: I like this pattern for testing. Would also be good to see some code to gast and gast to code tests if possible as well!

java/code_to_gast/java_expression.py
```diff
@@ -5,16 +5,36 @@
 gAST node of type funcCall
 '''
 def method_invocation_to_gast(node):
+
```
jackdavidweber, line 8: Nit: unnecessary blank line

java/code_to_gast/java_expression.py
```diff
@@ -5,16 +5,36 @@
...
     gast = {"type": "funcCall"}
     gast["args"] = java_router.node_to_gast(node.arguments)
     
     #TODO: change logic and add support for functions called on objects
     if node.qualifier == "System.out" and node.member == "println":
         gast["value"] = {"type": "logStatement"}
     else:
-        gast["value"] = {"type": "customStatement"}
+        # function not called on object
+        if not node.qualifier:
```
jackdavidweber, line 17: Nit: remove not and flip the if else. I find that it improves readability when if statements aren't negated

java/code_to_gast/java_expression.py
```diff
@@ -5,16 +5,36 @@
...
+            gast["value"] = {"type": "name", "value": node.member}
+        else:
+            object_list = (node.qualifier.split("."))
+            object_list.append(node.member)
+            gast["value"] = functionHelper(object_list)
+    return gast
+
+def functionHelper(object_list):
```
jackdavidweber, line 25: nit: camelCase instead of snake_case

jackdavidweber, line 25: Also would be great if you could add comment about functionality!

java/gast_to_code/gast_to_code_java.py
```diff
@@ -52,10 +52,17 @@ def handle_subscript(gast):
         pass
 
     def handle_name(gast):
-        return gast["value"]
+        ''' 
+        NOTE: some places store {"type": "name", "value": "s"} while others have
+        {"type": "name", "id" : "s"} in gAST but both get routed to this func.
+        We may want to re-evaluate gAST structure regarding funcs and vars
```
augerm, line 58: Can we create a taiga ticket for this? Since both have the same type it seems like we should stick to either `id` or `name`

java/code_to_gast/java_expression.py
```diff
@@ -12,9 +12,34 @@ def method_invocation_to_gast(node):
...
 
+'''
+Takes list of callees and members and from function called
+on object and translates into gAST node
+Ex: car.drive() -> {"type": "attribute", "id": "drive", "value": 
+{"type": "name", "id": "car"}}
+'''
+def list_to_attribute_value_node(object_list):
+    gast = {}
+    gast["type"] = "attribute"
```
augerm, line 32: Can we create a ticket to move all hardcoded strings like "attribute" to enums in the Java code? I would label these with a tag that says "Refactor"

java/gast_to_code/gast_to_code_java.py
Danish21: +1

java/code_to_gast/java_expression.py
Danish21: +1

PR #83, opened 2020-07-16
Bash conditionals

cjoshea9:
Decided that code_to_gast was too difficult with the limited tooling from these bash parsers, so with Michael, Jack, and Steph decided we just won't support bash as an input language, but only as an output language. 

I imagine that most use cases for this would be translating say python code to bash and not vice versa, so hopefully this makes sense for our users as well.

This PR adds gast_to_code for conditionals in bash. 

Note: there is no concept of booleans in bash so other things will need to be added to make these conditionals more effective and useful.


cjoshea9:
In the next PR I will get rid of all the code_to_gast with Bash if everyone agrees we should no longer support it as an input language.

bash/code_to_gast/bash_helpers.js
```diff
@@ -24,12 +24,8 @@ function array_to_gast(node) {
  * Word nodes are base cases such as strings and numbers
  */ 
 function word_to_gast(node) {
-    text = node["text"]
-    if (text === "true") {
-        return {"type": "bool", "value": 1}
-    } else if (text === "false") {
-        return {"type": "bool", "value": 0}
-    } else if (!isNaN(text)) {
```
cjoshea9, line 32: Gets rid of booleans as they don't exist

bash/code_to_gast/bash_helpers.js
```diff
@@ -24,12 +24,8 @@ function array_to_gast(node) {
...
+    const text = node["text"]
+    if (!isNaN(text)) {
```
augerm, line 28: Can we invert the if/else to avoid the double negation

bash/gast_to_code/bash_gast_to_code_converter.py
```diff
@@ -7,14 +7,27 @@ class BashGastToCodeConverter(AbstractGastToCodeConverter):
     pretty_name = "Bash"
    
     def handle_bool(gast):
-        if gast["value"] == 1:
-            return "true"
-        else:
-            return "false"
+        return "Bash does not support booleans"
```
augerm, line 10: I say we just return the strings "true" and "false" anyways since they end up working the same way.

main.py
```diff
@@ -26,6 +26,7 @@ def main(input_code, input_lang, output_lang):
     elif input_lang == "java":
         gast = java_main.java_to_gast(input_code)
     elif input_lang == "bash":
+        # TODO move to separate function
```
augerm, line 29: Can we move this now? :)

test/test_bash_primitives.py
```diff
@@ -2,12 +2,6 @@
 import main
 
 class TestBashPrimitives(unittest2.TestCase):
-    def test_boolean(self):
```
augerm, line 5: Don't forget to add this back if you add the air quote "boolean" logic back

bash/gast_to_code/bash_gast_to_code_converter.py
cjoshea9: Okay

PR #87, opened 2020-07-17
Bash arrays

jackdavidweber:
Implemented bash arrays. I found out that bash does not support printing arrays so I added in logic that prevents translation from `print[1,2]` to `echo (1, 2)`

### DIFF
--- a/bash/gast_to_code/bash_gast_to_code_converter.py
+++ b/bash/gast_to_code/bash_gast_to_code_converter.py
@@ -16,5 +16,13 @@ class BashGastToCodeConverter(AbstractGastToCodeConverter):
         return "echo"
     
     def handle_func_call(gast):
+        # bash has no way to directly print arrays. This logic returns an error for such behavior
+        # TODO: maybe we should run code after translation to make sure it compiles and throw errors if it does NOT
+        if gast["value"]["type"] == "logStatement" and general_helpers.arr_in_list(gast["args"]):
+            return "Error: direct translation not possible" # TODO: streamline error message as part of refactor
+
         return router.gast_to_code(gast["value"], "bash") + " " + router.gast_to_code(gast["args"], "bash")
+
+    def handle_arr(gast):
+        return "(" + router.gast_to_code(gast["elements"], "py") + ")"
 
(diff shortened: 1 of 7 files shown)


jackdavidweber:
TODO: throw error if nested arrays in bash

test/test_bash_primitives.py
```diff
@@ -13,6 +13,9 @@ def test_number(self):
         bash_code = '46'
         py_code = '46'
         self.assertEqual(bash_code, main.main(js_code, 'js', 'bash'))
+
+    def test_arr(self):
+        self.assertEqual('(1, 2)', main.main('[1, 2]', 'py', 'bash'))
```
swalsh15, line 18: Is [1, [2, 3]] possible in bash? If it is a test case for that would be great

swalsh15: nvm saw your update

bash/gast_to_code/bash_gast_to_code_converter.py
```diff
@@ -16,5 +16,16 @@ def handle_log_statement(gast):
         return "echo"
     
     def handle_func_call(gast):
+        # bash has no way to directly print arrays. This logic returns an error for such behavior
+        # TODO: maybe we should run code after translation to make sure it compiles and throw errors if it does NOT
```
augerm, line 20: I'd rather have this TODO in a taiga ticket. This seems applicable to all converters.

bash/gast_to_code/bash_gast_to_code_converter.py
```diff
@@ -16,5 +16,16 @@ def handle_log_statement(gast):
...
+        if gast["value"]["type"] == "logStatement" and general_helpers.arr_in_list(gast["args"]):
+            return "impossibleTranslationError: direct translation does not exist" # TODO: streamline error message as part of refactor
```
augerm, line 22: Nit: I'd move the #TODO about the return. And also for all these TODO's I would put #TODO(jackdavidweber) or #TODO(taigaTicket#)

bash/gast_to_code/bash_gast_to_code_converter.py
g-harel: I like the idea, but may not work  with unsuported code blocks in the output.

test/test_bash_primitives.py
```diff
@@ -13,6 +13,9 @@ def test_number(self):
...
```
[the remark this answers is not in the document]
jackdavidweber: Right now we are not testing any errors and since we are doing errors overhaul next week, I think that I am going to wait to add error test here

bash/gast_to_code/bash_gast_to_code_converter.py
jackdavidweber: Added taiga ticket- removing TODO

PR #89, opened 2020-07-17
Bash functions

swalsh15:
Add support for translations of custom function calls to bash. Includes functions with args that are primitives and variables.

### DIFF
--- a/shared/gast_to_code/gast_to_code_router.py
+++ b/shared/gast_to_code/gast_to_code_router.py
@@ -13,6 +13,9 @@ def gast_to_code(gast, out_lang, lvl=0):
     converter = ConverterRegistry.get_converter(out_lang)
     
     if type(gast) == list:
+        # bash has no comma between function arguments
+        if out_lang == "bash":
+            return general_helpers.list_helper(gast, out_lang, " ")
         return general_helpers.list_helper(gast, out_lang)
 
     # Primitives
--- a/test/test_custom_functions.py
+++ b/test/test_custom_functions.py
@@ -20,5 +20,8 @@ class TestCustomFunctions(unittest2.TestCase):
         self.assertEqual('car.honda.drive(1, 3)', main.main('car.honda.drive(1,3)', 'js', 'py'))
         self.assertEqual('car.ford.park(null)', main.main('car.ford.park(None)', 'py', 'js'))
 
+    def test_arg_varaible(self):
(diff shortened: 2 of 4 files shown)


shared/gast_to_code/gast_to_code_router.py
```diff
@@ -13,6 +13,9 @@ def gast_to_code(gast, out_lang, lvl=0):
     converter = ConverterRegistry.get_converter(out_lang)
     
     if type(gast) == list:
+        # bash has no comma between function arguments
+        if out_lang == "bash":
+            return general_helpers.list_helper(gast, out_lang, " ")
```
g-harel, line 18: Would be nice if this "list delimiter" was provided by the language instead of it needing to be known to the generic router.

shared/gast_to_code/gast_to_code_router.py
```diff
@@ -13,6 +13,9 @@ def gast_to_code(gast, out_lang, lvl=0):
     converter = ConverterRegistry.get_converter(out_lang)
     
     if type(gast) == list:
+        # bash has no comma between function arguments
+        if out_lang == "bash":
```
g-harel, line 17: Is there an enum of languages, this looks like the magic string problem there used to be in gast.

test/test_custom_functions.py
```diff
@@ -18,5 +20,8 @@ def test_object_attribute(self):
         self.assertEqual('car.honda.drive(1, 3)', main.main('car.honda.drive(1,3)', 'js', 'py'))
         self.assertEqual('car.ford.park(null)', main.main('car.ford.park(None)', 'py', 'js'))
 
+    #def test_arg_varaible(self):
```
g-harel, line 23: Does this work yet?

swalsh15: Just added a commit to support this test case @g-harel

jackdavidweber: Is the issue referenced in top level PR [comment](https://github.com/jackdavidweber/cjs_capstone/pull/89#issue-451332354) resolved? If so, maybe edit the comment to reflect it as otherwise it is slightly confusing

shared/gast_to_code/gast_to_code_router.py
Danish21: +1

shared/gast_to_code/gast_to_code_router.py
Danish21: +1

PR #92, opened 2020-07-17
Add and make accessible more information about each converter 

jackdavidweber:
Realized that the presence of beta languages in auto-detect was creating issues. We already had a [ticket](https://tree.taiga.io/project/jackdavidweber-cjs-capstone/us/142?no-milestone=1) for adding more information to the converters. This information could then be used to block beta languages from being used in auto-detect. The information was also supplied into the Get request

### GET Request Before ([PR](https://github.com/jackdavidweber/cjs_capstone/pull/81))
image

### GET Request After
image

### DIFF
--- a/app.py
+++ b/app.py
@@ -35,7 +35,7 @@ class Translate(Resource):
         # Gets non-beta (fully supported) languages for automatic detection
         fully_supported_lang_codes = ConverterRegistry.get_fully_supported_language_codes()
 
-        # automatic language detection (only fully supported languages)
+        # automatic language detection (only fully supported languages) TODO: fall back on Beta if all else fails
         i = 0
         while (output_code == "Error: did not compile") and (i < len(fully_supported_lang_codes)):
             response_input_lang = fully_supported_lang_codes[i]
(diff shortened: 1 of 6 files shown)


app.py
```diff
@@ -33,10 +32,13 @@ def post(self):
         output_code = main(input_code, input_lang, output_lang)
         response_input_lang = input_lang
 
-        # automatic language detection
+        # Gets non-beta (alpha) languages for automatic detection
```
g-harel, line 35: "Alpha" is usually even less stable than beta, is that what you mean?

app.py
```diff
@@ -33,10 +32,13 @@ def post(self):
...
+        alpha_lang_codes = ConverterRegistry.get_alpha_language_codes()
+
+        # automatic language detection (only alpha languages)
         i = 0
-        while (output_code == "Error: did not compile") and (i < len(self.lang_codes)):
-            response_input_lang = self.lang_codes[i]
+        while (output_code == "Error: did not compile") and (i < len(alpha_lang_codes)):
```
augerm: Can we fall back to the BETA languages if none of the stable languages compile?

PR #93, opened 2020-07-20
Support to use range function with one arg

swalsh15:
Support functionality of one argument loops in python to translate to other languages. 
```
for i in range(4):
    print(i)
```
Note: translations of loops TO python still use a 3 argument range function

### DIFF
--- a/python/code_to_gast/py_loop.py
+++ b/python/code_to_gast/py_loop.py
@@ -41,9 +41,14 @@ def for_range_statement_to_gast(node):
         gast["type"] = "forRangeStatement"
         ast.dump(node)
         args = node.iter.args
-        step_num = get_step_num(args)
-        start_num = arg_node_to_num(args[0])
-        end_num = arg_node_to_num(args[1])
+        if len(args) == 1:
+            step_num = 1
+            start_num = 0
+            end_num = arg_node_to_num(args[0])
+        else:
+            step_num = get_step_num(args)
+            start_num = arg_node_to_num(args[0])
+            end_num = arg_node_to_num(args[1])
 
         gast["init"] = for_range_statement_init_helper(node.target, start_num)
         gast["test"] = for_range_statement_test_helper(node.target, start_num, end_num)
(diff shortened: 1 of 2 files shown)


python/code_to_gast/py_loop.py
```diff
@@ -41,9 +41,14 @@ def for_range_statement_to_gast(node):
         gast["type"] = "forRangeStatement"
         ast.dump(node)
         args = node.iter.args
-        step_num = get_step_num(args)
```
jackdavidweber, line 44: per our live conversation, it makes a lot more sense to use the helper function pattern to handle only one argument rather than adding in a conditional to this level. Basically just add an if statement into `get_step_num` and then create `get_start_num` and `get_end_num` functions that do effectively the same thing. I think it will make it much clearer over time

python/code_to_gast/py_loop.py
```diff
@@ -41,9 +41,14 @@ def for_range_statement_to_gast(node):
...
-        start_num = arg_node_to_num(args[0])
-        end_num = arg_node_to_num(args[1])
+        if len(args) == 1:
```
augerm, line 44: Can we add a comment here or wherever this logic ends up living showing example code snippets that lead to each branch?

test/test_loops.py
```diff
@@ -56,6 +56,15 @@ def test_for_range_all_args_neg(self):
...
         self.assertEqual(js_code, main.main(py_code, 'py', 'js'))
 
+    def test_for_range_one_arg_loop(self):
+        ''' 
+        users can give range 1 args - translations give range 3 args 
+        hence js -> py is not supported for this translation 
+        '''
+        js_code = 'for (let i = 0; i < 5; i += 1) {\n\tconsole.log(i)\n}'
+        py_code = 'for i in range (5):\n\tprint(i)'
+        self.assertEqual(js_code, main.main(py_code, 'py', 'js'))
```
augerm, line 66: Can we add bash/java tests?

PR #94, opened 2020-07-20
Handling Errors: Prevent "internal server errors" by using Try/Catch

jackdavidweber:
Certain translations were causing the backend to send an "internal server error" to the frontend which looked pretty bad.

image
This PR attempts to capture nearly all possible errors using try/except and abort(). In the process, I also refactor main.py to make it more readable and easier to add error checking.

**NOTE:** the actual error messages that are returned should be considered temporary as they will shortly be [replaced by error objects](https://tree.taiga.io/project/jackdavidweber-cjs-capstone/us/166?no-milestone=1) as described in the [handling errors doc.](https://docs.google.com/document/d/1wccE4ECuD0vac6efbEdOXfkkDQD23zBN8bX1WyT4V30/edit?usp=sharing)
### Before

* `curl http://127.0.0.1:5000/ -d "input=5&in_lang=java&out_lang=py"`
    * `Traceback (most recent call last): File ....`
* `curl http://127.0.0.1:5000/ -d "input=for (i=0; i<10; i++){}&in_lang=js&out_lang=py"`
    * `Traceback (most recent call last): File ....`
### After
* `curl http://127.0.0.1:5000/ -d "input=5&in_lang=java&out_lang=py"`
    * `Feature not supported`
* `curl http://127.0.0.1:5000/ -d "input=for (i=0; i<10; i++){}&in_lang=js&out_lang=py"`
    * `Error: unable to convert generic ast to code`

### DIFF
--- a/java/code_to_gast/java_main.py
+++ b/java/code_to_gast/java_main.py
@@ -5,12 +5,18 @@ import java.code_to_gast.java_router as java_router
 takes js string and converts it to a generic AST
 """
 def java_to_gast(java_input):
-    # wrap input code in a class and function
-    class_wrapper = '''
-    class Test {{
-        public static void main(String[] args) {{
-            {java_input}
-        }}
-    }}'''.format(java_input = java_input)
-    input_ast = javalang.parse.parse(class_wrapper)
+    input_ast = ''
+
+    try:
+        # wrap input code in a class and function
+        class_wrapper = '''
+        class Test {{
+            public static void main(String[] args) {{
+                {java_input}
+            }}
+        }}'''.format(java_input = java_input)
+        input_ast = javalang.parse.parse(class_wrapper)
+    except:
+        "Error: code could not compile"
+    
     return java_router.node_to_gast(input_ast)
--- a/main.py
+++ b/main.py
@@ -15,41 +15,77 @@ output_lang: string representing output language of code
 return: string representing output code or error message
 """
 def main(input_code, input_lang, output_lang):
+    try:
+        # check arguments TODO(taiga#172): remove hard coded references
+        check_valid_args(input_code, input_lang, output_lang, ["js", "py", "java"], ["js", "py", "bash", "java"])
 
-    if input_lang == None and output_lang == None:
-        abort(400, "Error: must specify input and output languages")
+        # code to gast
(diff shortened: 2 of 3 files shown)


jackdavidweber:
@hosts can this be considered a _hot fix_ and thus merged directly into master rather than into development? @augerm

main.py
```diff
@@ -16,54 +15,77 @@
...
+        abort(400, "Error must specify valid input language. For example, js for javascript and py for python")
    
-    output_code = ""
-    if (type(gast) == str) :
+    if output_lang not in valid_output_langs:
+        abort(400, "Error must specify valid output language. For example, js for javascript and py for python")
+   
+
+def check_valid_gast(gast):
+    if (type(gast) == str) : # FIXME(swalsh15) why is this necessary??
```
jackdavidweber, line 67: @swalsh15

main.py
```diff
@@ -15,41 +15,77 @@
...
-    elif input_lang == "java":
-        gast = java_main.java_to_gast(input_code)
-    else:
-        #TODO: figure out hwo to do error messages
-        return "Error must specify input language. For example, js for javascript and py for python"
+        #gast to code
+        output_code = main_gast_to_code(gast, output_lang)
+
+        # analytics
+        main_analytics(input_code, output_code, input_lang, output_lang)
```
Danish21, line 30: Nit:
Describe the action in the name i.e. store_analytics to make it easier to understand what the function is doing without having to read the function body.

jackdavidweber: good point! Will address

java/code_to_gast/java_main.py
```diff
@@ -5,12 +5,18 @@
...
-    class_wrapper = '''
-    class Test {{
-        public static void main(String[] args) {{
-            {java_input}
-        }}
-    }}'''.format(java_input = java_input)
-    input_ast = javalang.parse.parse(class_wrapper)
+    input_ast = ''
+
+    try:
```
g-harel, line 10: Is this wrapping needed if the caller (`main`) is also catching and will ignore the original error?

jackdavidweber: From having played around with this- the first error that is caught is the one that is passed up. By having this here, we gain specificity about where the actual error is occurring.

PR #96, opened 2020-07-20
Add increment and decrement operators 

cjoshea9:
Add ++ and -- across all four supported languages and added for loop functionality in python to handle this new case

### DIFF
--- a/java/code_to_gast/java_router.py
+++ b/java/code_to_gast/java_router.py
@@ -19,7 +19,8 @@ def node_to_gast(node):
             return {"type": "bool", "value": 0}
         else:
             return "Unsupported prim"
-    
+    elif type(node) == str:
+        return {"type": "name", "value": node}
     elif type(node) == javalang.tree.MethodInvocation:
         return java_expression.method_invocation_to_gast(node)
     elif type(node) == javalang.tree.CompilationUnit:
@@ -45,6 +46,8 @@ def node_to_gast(node):
             return {"type": "error", "value": "unsupported"}
     elif type(node) == javalang.tree.VariableDeclarator:
         return java_assign.assign_to_gast(node)
+    elif type(node) == javalang.tree.MemberReference:
+        return java_assign.member_reference_to_gast(node)
     elif type(node) == javalang.tree.ArrayInitializer:
         return java_helpers.array_to_gast(node.initializers)
     else:   
--- a/python/code_to_gast/py_assign.py
+++ b/python/code_to_gast/py_assign.py
@@ -39,7 +39,7 @@ def augop_to_str(op):
         ast.Add: "+=", 
         ast.Mult: "*=", 
         ast.Div: "/=", 
-        ast.Sub: "-=",
+        ast.Sub: "-="
     }
     return op_to_str_map[type(op)]
 
--- a/python/code_to_gast/py_helpers.py
+++ b/python/code_to_gast/py_helpers.py
@@ -86,8 +86,9 @@ def pyop_to_str(op):
         ast.GtE: ">=",
         ast.Lt: "<",
         ast.LtE: "<=",
-        ast.Eq: "=="
-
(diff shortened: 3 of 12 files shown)


java/code_to_gast/java_router.py
```diff
@@ -19,7 +19,8 @@ def node_to_gast(node):
             return {"type": "bool", "value": 0}
         else:
             return "Unsupported prim"
-    
+    elif type(node) == str:
```
g-harel, line 22: What is this doing?

cjoshea9: Java doesn't have it's own type for variable names, so if we pass in just a string then we are going to assume its a variable name

g-harel: Interesting, so what was happening before this change?

python/code_to_gast/py_assign.py
```diff
@@ -39,7 +39,7 @@ def augop_to_str(op):
         ast.Add: "+=", 
         ast.Mult: "*=", 
         ast.Div: "/=", 
-        ast.Sub: "-=",
```
augerm, line 42: I think the google style guide will add this trailing comma back. At least that's what it does in JS :)

python/code_to_gast/py_helpers.py
```diff
@@ -86,8 +86,9 @@ def pyop_to_str(op):
         ast.GtE: ">=",
         ast.Lt: "<",
         ast.LtE: "<=",
-        ast.Eq: "=="
-
+        ast.Eq: "==",
+        ast.UAdd: "++",
```
augerm, line 90: I thought `++` and `--` aren't supported in python?

From my brief and very likely mistaken googling it looks like ast.UAdd might correspond with `+`. :)
https://www.programcreek.com/python/example/61006/ast.UAdd

python/gast_to_code/py_gast_to_code_converter.py
```diff
@@ -50,14 +50,28 @@ def handle_for_range(gast, lvl=0):
...
+                incrementor = "-" + str(incrementor_value)
+            elif incrementor_op == "+=":
+                incrementor = str(incrementor_value)
+            else:
+                incrementor = "unsupported update expression"
+        
+        # ++ or -- expression
         else:
-            incrementor = "unsupported update operation"
+            if incrementor_op == "++":
```
augerm, line 67: Will this convert `a++` --> `a += 1`?

test/test_loops.py
```diff
@@ -43,6 +43,21 @@ def test_for_inclusiverange_increment_two(self):
         input_js_code = 'for (let i = 0; i <= 10; i += 2) {\n\t5\n}'
         expected_py_code = 'for i in range (0, 12, 2):\n\t5'
         self.assertEqual(expected_py_code, main.main(input_js_code, 'js', 'py'))
+    
+    def test_for_with_update_expression_plus(self):
```
augerm, line 47: Can we add some bash and java tests?

java/code_to_gast/java_router.py
cjoshea9: In the variable assignment section it is doing the same thing just not going through the router

python/gast_to_code/py_gast_to_code_converter.py
cjoshea9: Currently for Python, we have it translating to `a++` to `++a`

test/test_loops.py
cjoshea9: We don't have for loops implemented in java and bash yet so can't test this use case there yet

python/code_to_gast/py_helpers.py
cjoshea9: After doing more research I was wrong that `++a` works in Python. It compiles on the Python ast parser but does not seem to be actually allowed in Python

PR #98, opened 2020-07-20
Yapf setup

jackdavidweber:
[yapf](https://opensource.google/projects/yapf) is a Google open source formatter for python files. In this PR, I did the following:

- Used yapf to automatically apply the google style guide to our backend repository while excluding directories that do not need to be formatted (i.e. experimental).
- Created a yaml file that is configured to automatically check if our code meets google style guides on every push.

The following command is used to auto-translate on all files locally (preferably before every push:
```
yapf -ir --verbose --style "google" .
```

### DIFF
--- a/.github/workflows/yapf_check.yml
+++ b/.github/workflows/yapf_check.yml
@@ -1,0 +1,12 @@
+name: YAPF Formatting Check
+on: [push]
+jobs:
+  formatting-check:
+    name: Formatting Check
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: run YAPF to test if python code is correctly formatted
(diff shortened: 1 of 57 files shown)


.github/workflows/yapf_check.yml
```diff
@@ -0,0 +1,12 @@
+name: YAPF Formatting Check
+on: [push]
+jobs:
+  formatting-check:
```
g-harel, line 4: Why wasn't this added to the existing workflow? It could replace the current lint step that we don't care about anymore.

jackdavidweber: good point- I originally couldn't figure out how to add to existing workflow but I figured it out

PR #99, opened 2020-07-21
Arrow/Anon functions

swalsh15:
Added support for arrow/anon functions in javascript and lambda functions. You can go from arrow or anon functions to lambda however lambda maps to arrow functions.

### DIFF
--- a/javascript/code_to_gast/js_expression.py
+++ b/javascript/code_to_gast/js_expression.py
@@ -59,6 +59,11 @@ def subscript_to_gast(node):
   gast["value"] = js_router.node_to_gast(node.object)
   return gast
 
+'''
+Takes arrow function or anon function and convert into a
+gast node. This node stores the same fields as a function
+node but doesn't store name
+'''
 def arrow_function_to_gast(node):
   gast = {}
   gast["type"] = "arrowExpression"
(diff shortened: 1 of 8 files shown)


javascript/code_to_gast/js_expression.py
```diff
@@ -58,3 +58,10 @@ def subscript_to_gast(node):
   gast["index"] = js_router.node_to_gast(node.property)
   gast["value"] = js_router.node_to_gast(node.object)
   return gast
+
+def arrow_function_to_gast(node):
+  gast = {}
+  gast["type"] = "arrowExpression"
```
jackdavidweber, line 64: make sure to add to example gast doc if you haven't already

javascript/code_to_gast/js_router.py
```diff
@@ -75,6 +75,10 @@ def node_to_gast(node):
         return js_helpers.return_statement_to_gast(node)
     elif node.type == "AssignmentPattern":
         return js_helpers.assign_pattern_to_gast(node)
+    elif node.type == "ArrowFunctionExpression":
```
jackdavidweber, line 78: What is the reasoning behind both arrowfunctionexpression and functionexpression doing the same thing?

swalsh15: The main difference between these two functions is how the keyword "this" is bound. I talked to Michael and he said for our functionality its fine to treat them the same. The arrow function is a newer feature which is why I prioritized it for py -> js

PR #100, opened 2020-07-21
Final yapf setup

jackdavidweber:
[yapf](https://opensource.google/projects/yapf) is a Google open source formatter for python files. In this PR, I did the following:

- Used yapf to automatically apply the google style guide to our backend repository while excluding directories that do not need to be formatted (i.e. experimental).
- Edited yaml file to automatically check if our code meets google style guides on every push/PR.

In order to run yapf on your branch:
```
pip install yapf
yapf -ir --verbose --style "google" .
```

### DIFF
--- a/app.py
+++ b/app.py
@@ -15,7 +15,9 @@ parser.add_argument('input')
 parser.add_argument('in_lang')
 parser.add_argument('out_lang')
 
+
 class Translate(Resource):
+
     def __init__(self):
         self.lang_object = ConverterRegistry.get_lang_dict()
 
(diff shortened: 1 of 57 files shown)


app.py
```diff
@@ -33,11 +35,13 @@ def post(self):
         response_input_lang = input_lang
 
         # Gets non-beta (fully supported) languages for automatic detection
-        fully_supported_lang_codes = ConverterRegistry.get_fully_supported_language_codes()
+        fully_supported_lang_codes = ConverterRegistry.get_fully_supported_language_codes(
+        )
```
cjoshea9, line 39: This seems to be wrong

java/code_to_gast/java_assign.py
```diff
@@ -1,20 +1,24 @@
...
     # var name stored as string but we don't want to return gast string node
     gast["varId"] = {"type": "name", "value": node.name}
     return gast
 
+
 """
 Handles increment and decrement operators
 """
+
+
```
cjoshea9, line 19: I don't think we want space between our doc string and function

app.py
Danish21: +1. This seems strange.

java/code_to_gast/java_assign.py
Danish21: +1. Is there a way to configure this?

g-harel: Python doc strings usually go _inside_ the function, I think that might be why it's getting confused.

app.py
g-harel: Agreed, seems like a quirk of setting a fixed line cutoff. Not sure there's much to do about it, maybe could move the `.get_fully_...` to the next line.

PR #101, opened 2020-07-21
Handling errors refactor

jackdavidweber:
First step was to fix all of the tests so that anything that was expecting an output from main or gast to code would test an object rather than a string. Since we have over 130 tests, I decided to use regex rather than manually fixing all of the tests.

### Regex for tests using main.main:
```
main.main\((.*)\)\) # find
main.main($1)['translation']['output_code']) # replace
```

### Regex for tests using gtc.gast_to_code:
```
gtc.gast_to_code\((.*)\)\) # find
gtc.gast_to_code($1)['translation']['output_code']) # replace
```

### DIFF
--- a/debug.py
+++ b/debug.py
@@ -1,3 +1,7 @@
-import bash.code_to_gast.bash_main as bash_main
+import main
+from bootstrap import bootstrap
 
-print(bash_main.bash_to_gast('true'))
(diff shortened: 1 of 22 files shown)


g-harel:
Woot, formatter presubmit is working!

PR #102, opened 2020-07-21
Java for loops, while loops, augmented assignment, binary operations

cjoshea9:
Adds for loops, while loops, augmented assignment and none type. Augmented assignment should be rexamined as it doesn't cover every use case yet

### DIFF
--- a/java/code_to_gast/java_assign.py
+++ b/java/code_to_gast/java_assign.py
@@ -26,7 +26,7 @@ def member_reference_to_gast(node):
 
 def aug_assign_to_gast(node):
     """
-    Handles augmented assignemnt in java but not the incrementor and decrementor operations
+    Handles augmented assignment in java but not the incrementor and decrementor operations
     """
     gast = {"type": "augAssign"}
     gast["left"] = java_router.node_to_gast(node.expressionl.member)
--- a/java/code_to_gast/java_assign.py
+++ b/java/code_to_gast/java_assign.py
@@ -22,3 +22,14 @@ def member_reference_to_gast(node):
     gast["left"] = java_router.node_to_gast(node.member)
     gast["op"] = node.postfix_operators[0]
     return gast
+
+
+def aug_assign_to_gast(node):
+    """
+    Handles augmented assignemnt in java but not the incrementor and decrementor operations
+    """
+    gast = {"type": "augAssign"}
+    gast["left"] = java_router.node_to_gast(node.expressionl.member)
+    gast["op"] = node.type
+    gast["right"] = java_router.node_to_gast(node.value)
+    return gast
--- a/java/code_to_gast/java_helpers.py
+++ b/java/code_to_gast/java_helpers.py
@@ -25,3 +25,11 @@ def array_to_gast(node):
 
     gast["elements"] = gast_list
     return gast
+
+
+def bin_op_to_gast(node):
+    gast = {"type": "binOp"}
+    gast["left"] = java_router.node_to_gast(node.operandl.member)
+    gast["op"] = node.operator
+    gast["right"] = java_router.node_to_gast(node.operandr)
+    return gast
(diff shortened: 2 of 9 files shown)


java/code_to_gast/java_main.py
```diff
@@ -17,6 +17,7 @@ class Test {{
             }}
         }}'''.format(java_input=java_input)
         input_ast = javalang.parse.parse(class_wrapper)
+        print(input_ast)
```
Danish21: Remove?

java/code_to_gast/java_assign.py
```diff
@@ -22,3 +22,14 @@ def member_reference_to_gast(node):
     gast["left"] = java_router.node_to_gast(node.member)
     gast["op"] = node.postfix_operators[0]
     return gast
+
+
+def aug_assign_to_gast(node):
+    """
+    Handles augmented assignemnt in java but not the incrementor and decrementor operations
```
Danish21, line 29: Typo: s/assignemnt/assignment

java/gast_to_code/gast_to_code_java.py
```diff
@@ -36,16 +36,42 @@ def handle_if(gast, lvl=0):
...
+        out = 'while (' + test + ') {' + body_indent + body + closing_brace_indent + "}"
+        return out
 
     def handle_for_range(gast, lvl=0):
-        pass
+        loop_init = router.gast_to_code(gast["init"], "java")
+        loop_test = router.gast_to_code(gast["test"], "java")
+        loop_update = router.gast_to_code(gast["update"], "java")
+
+        body_indent = "\n\t" + "\t" * lvl
```
Danish21, line 57: Consider creating helper functions for body_indent and closing_brace_indent since they are used in multiple places.

test/test_loops.py
```diff
@@ -84,6 +105,24 @@ def test_for_range_one_arg_loop(self):
...
+        self.assertEqual(js_code, main.main(py_code, "py", "js"))
+
+    def test_forOf_with_java(self):
+        js_code = 'for (elem of arr) {\n\t5\n}'
+        py_code = 'for elem in arr:\n\t5'
+        java_code = 'for (int elem : arr) {\n\t5;\n}'
+
+        self.assertEqual(js_code, main.main(py_code, "py", "js"))
+        self.assertEqual(py_code, main.main(js_code, "js", "py"))
+        # Without context can't determine what type the elements in the array are,
```
Danish21, line 121: Maybe we used just a GenericType i.e. a fake class to indicate that we do not know the type?

java/code_to_gast/java_helpers.py
```diff
@@ -25,3 +25,23 @@ def array_to_gast(node):
...
+    gast["op"] = node.operator
+    gast["right"] = java_router.node_to_gast(node.operandr)
+    return gast
+
+
+def int_to_gast(node):
+    """
+    Handles int to gast for positive and negative whole numbers
+    """
+    if len(node.prefix_operators) > 0 and node.prefix_operators[0] == "-":
```
jackdavidweber, line 45: Does the len(node.prefix_operators) > 0 do anything? In python, both objects and arrays have length so you might need to make sure it is an array as well since that is your intended purpose.
```
d = {'a':1, 'b':2, 'c':3}
print(len(d)) # ==> 3
```

java/code_to_gast/java_router.py
```diff
@@ -47,10 +52,18 @@ def node_to_gast(node):
             return {"type": "error", "value": "unsupported"}
     elif type(node) == javalang.tree.VariableDeclarator:
         return java_assign.assign_to_gast(node)
+    elif type(node) == javalang.tree.VariableDeclaration:
+        return node_to_gast(node.declarators[0])
```
jackdavidweber, line 56: Is there a way to generalize this so that we can just pass in node?

PR #103, opened 2020-07-22
Functionality for function declarations

swalsh15:
This adds support for translating functions to and from java. This code includes two wrappers - one with a class and one with a class and a function to allow the user to type in statements or functions in java. We are using strings temporarily to represent unknown information in functions such as return type and argument type.
If the user just types in statements (ie print, var declarations, etc) they will receive a translation of just the equivalent statements in java. 
If the user types a function they will receive the translated function in a java class. 
If the user includes both a function and statements outside the function we will return the translated function and the statements inside the java main function

Example translation js -> java 
Input:
```
console.log(1);
```
Output:
```
System.out.println(1);
```
Input:
```
function add(x) {
    1;
}
```
Output:
```
class Test {
    public unknown unknown add(customType x) {
        1; 
    }
}
```
Input:
```
function x() {
    1;
}
console.log(2);
```
Output:
```
class Test {
    public unknown unknown x() {
        1;
    }
    public static void main(String[] args) {
        System.out.println(2);
    }
}
```


java/code_to_gast/java_expression.py
```diff
@@ -53,6 +54,25 @@ def list_to_attribute_value_node(object_list):
...
 def class_declaration_to_gast(node):
     gast = {"type": "root"}
-    # only support for one function currently
-    gast["body"] = java_router.node_to_gast(node.body[0])
+    '''
+    If the method in class is main function only translate that method and ignore function header
+    We are assuming user only wants body of function translated
+    '''
+    if type(node.body[0]) == javalang.tree.MethodDeclaration and node.body[
+            0].name == "artifical_wrapper_WkVHC":
```
cjoshea9, line 62: What does the "artificial_wrapper_WkVHC" mean?

cjoshea9: Maybe add a comment to explain

main.py
```diff
@@ -16,25 +16,25 @@
 
 
 def main(input_code, input_lang, output_lang):
-    try:
-        # check arguments TODO(taiga#172): remove hard coded references
-        check_valid_args(input_code, input_lang, output_lang,
-                         ["js", "py", "java"], ["js", "py", "bash", "java"])
+    #try:
```
cjoshea9, line 19: Should you add these try catches back in now?

java/code_to_gast/java_expression.py
swalsh15: Ill add a comment - this is a function name that signifies that the user just typed in commands ie System.out.println(). We need to store the user's input in a class and function to compile in java. We don't store this function as a function declaration in the gast we just store the body of this function so only the commands are translated.

bash/gast_to_code/bash_gast_to_code_converter.py
```diff
@@ -28,7 +28,7 @@ def handle_arr(gast):
         if general_helpers.arr_in_list(gast["elements"]):
             return "impossibleTranslationError: direct translation does not exist"  # TODO: streamline error message as part of refactor
 
-        return "(" + router.gast_to_code(gast["elements"], "py") + ")"
+        return "(" + router.gast_to_code(gast["elements"], "bash") + ")"
```
jackdavidweber, line 31: why is this in this PR?

java/code_to_gast/java_expression.py
```diff
@@ -53,6 +54,25 @@ def list_to_attribute_value_node(object_list):
...
+
+def function_delcaration_to_gast(node):
+    gast = {"type": "functionDeclaration"}
+    gast["params"] = java_router.node_to_gast(node.parameters)
+    gast["id"] = {"type": "name", "value": node.name}
+    gast["body"] = java_router.node_to_gast(node.body)
     return gast
+
+
+def formal_parameter_to_gast(node):
```
jackdavidweber, line 77: add comment- what is a formal vs. informal parameter? Might not be important but if it isn't then just specify as such

java/code_to_gast/java_main.py
```diff
@@ -6,18 +6,29 @@
 
 
 def java_to_gast(java_input):
-    input_ast = ''
+    '''
+    Input will be wrapped in a class and main function if no functions are present
+    If functions are present it will be wrapped in a class
```
jackdavidweber, line 11: good job with this comment!

java/gast_to_code/gast_to_code_java.py
```diff
@@ -139,7 +139,23 @@ def handle_unary_op(gast):
         pass
 
     def handle_function_declaration(gast, lvl=0):
-        pass
+        name = router.gast_to_code(gast["id"], "java")
```
jackdavidweber, line 142: prob should add comments explaining what we are doing here.

java/gast_to_code/gast_to_code_java.py
```diff
@@ -149,3 +165,29 @@ def handle_assign_pattern(gast):
 
     def handle_arr(gast):
         return "{" + router.gast_to_code(gast["elements"], "java") + "}"
+
+    def handle_root(gast):
```
jackdavidweber, line 169: Also could use some comments since this is not very intuitively clear. Feel like if I were to come back in a month to look at this I would be confused even though i helped write it.

test/test_function_declarations.py
```diff
@@ -55,6 +58,31 @@ def test_multiple_default_vals(self):
         self.assertEqual(py_code, main.main(js_code, 'js', 'py'))
         self.assertEqual(js_code, main.main(py_code, 'py', 'js'))
 
+    def test_java_simple(self):
```
jackdavidweber, line 61: Nit: more specific test name than "simple"

debug.py
```diff
@@ -1,3 +1,10 @@
-import bash.code_to_gast.bash_main as bash_main
+import main as main
+from bootstrap import bootstrap
 
-print(bash_main.bash_to_gast('true'))
\ No newline at end of file
+# Register all languages before testing can begin
+bootstrap()
+
+print(
```
Danish21, line 7: Do we want to submit changes to debug.py?

java/code_to_gast/java_expression.py
Danish21: Pull this into a top level constant and add a comment there.

java/code_to_gast/java_expression.py
Danish21: +1

java/code_to_gast/java_main.py
Danish21: +1

java/code_to_gast/java_main.py
```diff
@@ -6,18 +6,29 @@
...
 def java_to_gast(java_input):
-    input_ast = ''
+    '''
+    Input will be wrapped in a class and main function if no functions are present
+    If functions are present it will be wrapped in a class 
+    If a class declaration is given an error will be thrown 
+    '''
+    class_main_wrapped = '''
+    class Test {{
+        public static void artifical_wrapper_WkVHC(String[] args) {{
```
Danish21, line 16: Why suffix with WkVHC?

java/gast_to_code/gast_to_code_java.py
Danish21: +1

java/gast_to_code/gast_to_code_java.py
```diff
@@ -139,7 +139,23 @@ def handle_unary_op(gast):
         pass
 
     def handle_function_declaration(gast, lvl=0):
-        pass
+        name = router.gast_to_code(gast["id"], "java")
+        if len(gast["params"]) != 0:
+            args = "customType "
```
Danish21, line 144: s/customType/CustomType since Java classes should be capitalized.

bash/gast_to_code/bash_gast_to_code_converter.py
swalsh15: this was causing test failures - think its a bug from another PR

debug.py
swalsh15: No - committed this accidentally

java/code_to_gast/java_main.py
swalsh15: just tried to pick a random string of characters to decrease the chance someone accidentally names their function this keyword

bash/gast_to_code/bash_gast_to_code_converter.py
```diff
@@ -51,7 +51,7 @@ def handle_if(gast, lvl=0):
 
     def handle_var_assign(gast):
         value = router.gast_to_code(gast["varValue"], "bash")
-        return router.gast_to_code(gast["varId"], "bash") + "=" + value
+        return router.gast_to_code(gast["varId"], "bash") + " = " + value
```
jackdavidweber, line 54: FYI: the reason that I did not add spaces originally is bc bash treats spaces in a strange way. Not sure what is the right way but wanted to give you a heads up!

PR #104, opened 2020-07-22
Change POST request to have info about detect language

cjoshea9:


bootstrap.py
```diff
@@ -6,7 +6,7 @@
 
 
 def bootstrap():
-    ConverterRegistry.register(PyGastToCodeConverter, "py")
     ConverterRegistry.register(JsGastToCodeConverter, "js")
+    ConverterRegistry.register(PyGastToCodeConverter, "py")
```
g-harel, line 10: What's this about?

app.py
```diff
@@ -31,20 +31,22 @@ def post(self):
...
+        if input_lang == "auto":
+            # Gets non-beta (fully supported) languages for automatic detection
+            fully_supported_lang_codes = ConverterRegistry.get_fully_supported_language_codes(
+            )
+            # automatic language detection (only fully supported languages) TODO: fall back on Beta if all else fails
+            i = 0
+            while i < len(fully_supported_lang_codes):
+                response_input_lang = fully_supported_lang_codes[i]
+                output_code = main(input_code, response_input_lang, output_lang)
+                i += 1
```
g-harel, line 43: Nit: Change to a fancier loop construct that doesn't require dealing with `i`.

bootstrap.py
cjoshea9: The order these are placed in the bootstrap file is the order they are taken and looped through in automatic language detection

app.py
```diff
@@ -31,20 +31,20 @@ def post(self):
...
+        if input_lang == "auto":
+            # Gets non-beta (fully supported) languages for automatic detection
+            fully_supported_lang_codes = ConverterRegistry.get_fully_supported_language_codes(
+            )
+            # automatic language detection (only fully supported languages) TODO: fall back on Beta if all else fails
+            for lang in fully_supported_lang_codes:
+                response_input_lang = lang
+                output_code = main(input_code, response_input_lang, output_lang)
+
+                if "Error" not in output_code:
```
swalsh15, line 43: will it ever be the case no languages compile and is there an error message for that?

bootstrap.py
augerm: Could we set a priority on the converter? LOW, MEDIUM, HIGH and use that to determine order?

augerm: or maybe change the call signature to `ConverterRegistry.register(PyGastToCodeConverter, "py", Priority.Low)` and let the converter registry handle this.

augerm: Just want to avoid the implicit prioritization here that might be confusing to someone later on when adding a new language.

app.py
cjoshea9: Yes, once the last language doesn't compile, it will just return "Error: could not compile"

PR #106, opened 2020-07-24
Semicolon update

cjoshea9:
Since we are printing all the code from gast_to_code, we know what characters can end lines, so although this may not be the most clean approach, it should produce readable and correct code for our users.


shared/gast_to_code/general_helpers.py
```diff
@@ -60,5 +60,7 @@ def java_linter(output_code):
     output_code = output_code.replace("{;", "{")
     output_code = output_code.replace("};", "}")
     output_code = output_code.replace("+;", "+")
+    output_code = output_code.replace("\t;", "\t")
+    output_code = output_code.replace(".;", ".")
```
augerm, line 64: Ha for some funny breaking cases, console.log() with emojis would probably break some of these cases.

However... Given the time constraints I think this is okay for now. But this should definitely have a TODO(<taiga-ticket-#>)

PR #114, opened 2020-07-28
Hot fix: Restructure gast to remove {type: name} having either id or value

swalsh15:
Our old gast had instances of {"type": "name", "value": "x"} and {"type": "name", "id": "x"}. I removed all instances of {"type": "name", "id": "x"} and changed id to value and updated the GAST contract doc. It turns out a lot of examples in the contract used id but the code_to_gast used value. Note: there are places that use id however they are {"type": "attribute", "id": .... }


swalsh15:
Quick fix: fixes an issue where java functions throw server errors when being translated to other languages. For example `system.f()` in java to python

PR #123, opened 2020-07-31
Fix testing

cjoshea9:
Created a testing matrix to do our language to language testing, then will create language specific files for the gast_to_code and code_to_gast tests. The code coverage is much higher now that every test goes through the testing matrix of all languages. Created a class called Unittest for each integration test and put all gasts in one shared file. 

This should make adding tests for new languages much quicker and cleaner. 

Apologies in advance for the large PR.


cjoshea9:
I also recommend looking at the changes all at once instead of commit by commit, because I continuously wrote over the same file as I changed my way of thinking about how to structure everything

test/Unittest.py
```diff
@@ -0,0 +1,19 @@
...
+        self.is_input = is_input
+        self.is_output = is_output
+
+    def get_lang(self):
+        return self.language
+
+    def get_code(self):
+        return self.code
+
+    def get_input(self):
```
g-harel, line 15: Nit: Would prefer `get_is_input`, this read as "get the input code" when I read it the first time.

test/matrix.py
```diff
@@ -0,0 +1,19 @@
...
+    output_langs = []
+
+    for obj in code_objects:
+        if obj.get_input():
+            input_langs.append(obj)
+        if obj.get_output():
+            output_langs.append(obj)
+
+    for input_lang in input_langs:
+        for output_lang in output_langs:
```
g-harel, line 15: Is it intended for a language that is both input and output to get compared to itself?

test/test_conditionals.py
```diff
@@ -1,53 +1,67 @@
...
-        bash_code = 'if [[ 1 ]]; then\n\techo "1 is true"\nelse\n\techo "1 is NOT true"\nfi'
-        java_code = 'if (1) {\n\tSystem.out.println("1 is true");\n} else {\n\tSystem.out.println("1 is NOT true");\n}'
-        self.assertEqual(py_code, translate.translate(js_code, 'js', 'py'))
-        self.assertEqual(js_code, translate.translate(py_code, 'py', 'js'))
-        self.assertEqual(bash_code, translate.translate(py_code, 'py', 'bash'))
-        self.assertEqual(bash_code, translate.translate(js_code, 'js', 'bash'))
-        self.assertEqual(java_code, translate.translate(js_code, 'js', 'java'))
-        self.assertEqual(py_code, translate.translate(java_code, 'java', 'py'))
+        js_code = Unittest(
+            'if (1) {\n\tconsole.log("1 is true")\n} else {\n\tconsole.log("1 is NOT true")\n}',
```
g-harel, line 20: Nit: Multi-line strings would be more readable.

cjoshea9: The way this was broken up was done by the Google linter, so I did not want to mess with that.

PR #128, opened 2020-07-31
Remove static language refs

jackdavidweber:
In pursuit of the goal of modularization of the languages, this PR attempts to remove static references to languages. The goal, once again, is to be able to add new languages by just editing the language-specific folder and the bootstrap file. 

This PR removes nearly all static references to languages in translate.py and main.py

#### Questionable Design Decision
Right now, the only way to access languages is through the converter. In order to remove static references to languages when going from code to gast, I had to add code_to_gast methods to all of the converters. This is confusing though since our converters are inside the gast to code directories for each language.

#### Recommended Solution
Converters should be pulled out of gast to code and put into the top level of the language directory. That way, converters can call gast to code and code to gast functions. It also allows the error handlers to be more generalized (right now error handlers exist in gast to code). I have created [taiga#206](https://tree.taiga.io/project/jackdavidweber-cjs-capstone/us/206?no-milestone=1) to address this issue.


translate.py
```diff
@@ -64,13 +34,10 @@ def is_valid_gast(gast):
 
 
 def gast_to_code_caller(gast, output_lang, error_handler):
-    # try:
     output_code = gtc.gast_to_code(gast, output_lang)
 
+    # TODO: remove last static reference to language
```
jackdavidweber, line 39: @cjoshea9 can you resolve this? Will prob require adding a field to the converter

jackdavidweber: [Taiga#207](https://tree.taiga.io/project/jackdavidweber-cjs-capstone/us/207?no-milestone=1)

PR #129, opened 2020-08-05
Post Processing to remove static refs

jackdavidweber:
Removed java static reference from translate by introducing idea of post processing to each converter.

### DIFF
--- a/java/gast_to_code/java_helpers.py
+++ b/java/gast_to_code/java_helpers.py
@@ -67,3 +67,26 @@ def java_node_list_helper(gast_list,
                 out += csv_delimiter
 
     return out
+
+
+def java_linter(output_code):
+    """
+    Add semicolons to the end of java lines
+    """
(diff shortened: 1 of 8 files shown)


java/gast_to_code/java_helpers.py
```diff
@@ -67,3 +67,26 @@ def java_node_list_helper(gast_list,
                 out += csv_delimiter
 
     return out
+
+
+def java_linter(output_code):
```
jackdavidweber, line 72: this was simply moved from general helpers to java helpers

shared/gast_to_code/general_helpers.py
```diff
@@ -37,26 +37,3 @@ def arr_in_list(gast_list):
             return True
 
     return False
-
-
-def java_linter(output_code):
```
jackdavidweber, line 42: moved to java_helpers

